PR #31, opened 2023-10-26
OPRUN-3075: Update to Upstream v0.7.0

dtfranz:


grokspawn:
We have a couple of "UPSTREAM: <drop>" commits in here which only manipulate the downstream-only elements of the repo.  Is it intentional that on next sync we would drop them and generate 2 new ones (manifests + vendors)?

dtfranz:
> We have a couple of "UPSTREAM: \<drop\>" commits in here which only manipulate the downstream-only elements of the repo. Is it intentional that on next sync we would drop them and generate 2 new ones (manifests + vendors)?

Yup! That's exactly what I did with the `<drop>` commits that were in the git history before.

PR #45, opened 2023-11-27
UPSTREAM: <carry>: Add commitchecker config

m1kola:
Making use of https://github.com/openshift/build-machinery-go/pull/82


m1kola:
It looks like we won't need it after https://github.com/openshift/operator-framework-tooling/pull/9

PR #52, opened 2023-12-11
OCPBUGS-22616: [release-4.14] Bump github.com/operator-framework/operator-registry

ankitathomas:


ncdc:
/hold

There are a lot of dependency updates in here. Is it possible to limit to solely operator-registry?

ankitathomas:
After looking through the library affected by the CVE, we aren't affected since we don't use the affected `otelhttp.NewHandler` method. This is an indirect dependency through operator-registry's CRD validation, which operator-controller does not use.

ncdc:
@ankitathomas nice!!!

PR #58, opened 2023-12-18
OCPBUGS-22616: [release-4.14] Bump go.opentelemetry.io dependencies

ankitathomas:


ncdc:
@ankitathomas please update your commit message to start with `UPSTREAM: `. Is there a corresponding PR to operator-framework/operator-controller that makes/made this change? We'd want the prefix to be `UPSTREAM: 1234` with 1234 being replaced with whatever the upstream PR was. If there isn't an upstream PR, we should probably have one?

m1kola (approved):
> Is there a corresponding PR to operator-framework/operator-controller that makes/made this change? We'd want the prefix to be `UPSTREAM: 1234` with 1234 being replaced with whatever the upstream PR was. If there isn't an upstream PR, we should probably have one?

We have https://github.com/operator-framework/operator-controller/pull/557 which bumps `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp` amongst other things. However it doesn't look like we will be able to cherry-pick the relevant commit into 4.14 downstream branch.

As far as I understand, in this case it should be a `<carry>` (as it is right now). So it looks good to me.

/lgtm

awgreene (approved):
/approve

PR #148, opened 2024-09-23
NO-ISSUE: UPSTREAM: <carry>: Properly copy and call kustomize

tmshort:
This fixes the calling of kustomize similar to https://github.com/openshift/operator-framework-catalogd/pull/63


perdasilva (approved):
/lgtm
/approve

PR #151, opened 2024-09-26
OPRUN-3557: Add test-e2e target for downstream makefile

dtfranz:
Target is set to always pass while we work on getting the tests to pass downstream. Adding this target allows us to make e2e test changes without having to make PRs against the `openshift/release` repo each time.

### DIFF
--- a/openshift/Makefile
+++ b/openshift/Makefile
@@ -14,3 +14,7 @@ manifests: $(KUSTOMIZE) $(YQ)
 .PHONY: verify-manifests
 verify-manifests: manifests
 	git diff --exit-code
+
+.PHONY: test-e2e
+test-e2e: ## Run the e2e tests. TODO: stub until tests are working downstream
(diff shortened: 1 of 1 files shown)


openshift/Makefile
```diff
@@ -14,3 +14,6 @@ manifests: $(KUSTOMIZE) $(YQ)
 .PHONY: verify-manifests
 verify-manifests: manifests
 	git diff --exit-code
+
+.PHONY: test-e2e
+test-e2e: /bin/true ## Run the e2e tests. TODO: stub until tests are working downstream
```
tmshort, line 19: The `/bin/true` shouyld be part of the recipe, indented on a separate line.

tmshort:
Use OPRUN-3557

openshift/Makefile
dtfranz: Woops!

PR #155, opened 2024-09-27
NO-ISSUE: Add downstream verify makefile target

dtfranz:


openshift/Makefile
```diff
@@ -7,6 +7,12 @@ include $(addprefix $(DIR)/vendor/github.com/openshift/build-machinery-go/make/,
 
 include $(DIR)/.bingo/Variables.mk
 
+.PHONY: verify
+verify: ## Run downstream-specific verify
+	$(MAKE) tidy fmt vet generate -C $(DIR)/../
+	$(MAKE) manifests
```
tmshort, line 13: This could just be a dependency for `verify`, so no need for another make line... unless tidy, etc need to be run first.

joelanford: They are run out of the upstream Makefile, so they aren't available here unless we `include` them (and including them might cause more trouble)

tmshort: In this case, it's running the `verify` target in _this_ makefile (no `-C` option or `cd` command)

dtfranz: That's right, it's running the downstream manifests target in this file. We do need to run tidy + fmt + vet + generate first though, so I think it needs to be kept in this order.

PR #191, opened 2024-11-15
OPRUN-3587: Enable downstream e2e

dtfranz:
Removes the fake e2e `/bin/true` stub and enables the upstream e2e. Also adds an additional script to deploy the e2e bundle/catalog registry into the oc ci environment.
~~Requires:
https://github.com/openshift/release/pull/58887
https://github.com/openshift/operator-framework-operator-controller/pull/192
https://github.com/openshift/release/pull/58802~~

### DIFF
--- a/openshift/registry.Dockerfile
+++ b/openshift/registry.Dockerfile
@@ -11,6 +11,7 @@ USER 1001
 COPY --from=builder /build/registry /registry
 COPY --from=builder /build/push /push
 COPY openshift/manifests /openshift/manifests
+COPY testdata/images /images
 
 LABEL io.k8s.display-name="OpenShift Operator Lifecycle Manager Operator Controller E2E Registry" \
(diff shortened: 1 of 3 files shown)


dtfranz:
There was a failure deploying the image registry, I'm looking into it now.
/hold

openshift/registry.Dockerfile
```diff
@@ -11,6 +11,7 @@ USER 1001
 COPY --from=builder /build/registry /registry
 COPY --from=builder /build/push /push
 COPY openshift/manifests /openshift/manifests
+COPY testdata/images /images
```
grokspawn, line 14: I think this file contributes to ose-operator-registry IIRC which serves as the foundation for all current RH catalog images.  
I don't think that we should be introducing test data to all catalog images.

grokspawn:
Similarity in naming made me think for a second that the dockerfile here was the same as used to build ose-operator-registry, but it looks like this is not the case.  

/lgtm
/approve

PR #197, opened 2024-11-20
OCPBUGS-44813: UPSTREAM: <drop>: catalog cache: retry cache population when cache contains an error

joelanford:
Temporary drop patch for https://github.com/operator-framework/operator-controller/pull/1489

### DIFF
--- a/internal/catalogmetadata/cache/cache.go
+++ b/internal/catalogmetadata/cache/cache.go
@@ -60,19 +60,6 @@ func (fsc *filesystemCache) Put(catalogName, resolvedRef string, source io.Reade
 	fsc.mutex.Lock()
 	defer fsc.mutex.Unlock()
 
-	// make sure we only write if this info hasn't been updated
-	// by another thread. The check here, if multiple threads are
-	// updating this, has no way to tell if the current ref is the
-	// newest possible ref. If another thread has already updated
-	// this to be the same value, skip the write logic and return
-	// the cached contents.
-	if cache, err := fsc.get(catalogName, resolvedRef); err == nil && cache != nil {
-		// We only return here if the was no error during
-		// the previous (likely concurrent) cache population attempt.
-		// If there was an error - we want to try and populate the cache again.
-		return cache, nil
-	}
-
 	var cacheFS fs.FS
 	if errToCache == nil {
 		cacheFS, errToCache = fsc.writeFS(catalogName, source)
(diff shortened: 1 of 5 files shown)


everettraven (approved):
/approve
/lgtm

internal/catalogmetadata/cache/cache.go
```diff
@@ -60,19 +60,6 @@ func (fsc *filesystemCache) Put(catalogName, resolvedRef string, source io.Reade
...
-	// newest possible ref. If another thread has already updated
-	// this to be the same value, skip the write logic and return
-	// the cached contents.
-	if cache, err := fsc.get(catalogName, resolvedRef); err == nil && cache != nil {
-		// We only return here if the was no error during
-		// the previous (likely concurrent) cache population attempt.
-		// If there was an error - we want to try and populate the cache again.
-		return cache, nil
-	}
-
```
joelanford, line 75: Multiple threads can't be simultaneously `Put`-ing because we hold a write lock for the duration of the `Put` call. This is leftover from a time when we performed this read with a read lock and then proceeded to a write lock.

We also want to do the expected thing and always update the cache with the latest `Put`, which ensures eventual consistency when combined with the updates made to the reconciler.

internal/catalogmetadata/cache/cache_test.go
```diff
@@ -96,12 +96,10 @@ func TestFilesystemCachePutAndGet(t *testing.T) {
...
-	actualFSPut, err = c.Put(catalogName, resolvedRef1, nil, errors.New("fake put error"))
-	// Errors do not override previously successfully populated cache
-	require.NoError(t, err)
-	require.NotNil(t, actualFSPut)
-	assert.NoError(t, equalFilesystems(defaultFS(), actualFSPut))
-	assert.NoError(t, equalFilesystems(actualFSPut, actualFSGet))
+	actualFSPut, err = c.Put(catalogName, resolvedRef1, nil, errors.New("fake v1 put error"))
+	// Errors for an existing resolvedRef should override previously successfully populated cache
+	assert.Equal(t, err, errors.New("fake v1 put error"))
+	assert.Nil(t, actualFSPut)
```
joelanford, line 102: With the changes in `internal/catalogmetadata/cache/cache.go`, we now expect `Put` to always update the cache, even for the same resolvedRef.

joelanford:
Strange.. I noticed that the same CI namespace "ci-op-0z4bhb87" was being re-used between test runs, and the original build namespace and build pod stuck around, causing subsequent test runs to somehow rediscover it and then immediately fail because they immediately detected that same pod that had been pending for over an hour.

I went ahead and manually deleted that namespace, so let's see if this latest test attempt succeeds. 🤞

joelanford:
No use. Seems we are affected by https://redhat-internal.slack.com/archives/CBN38N3MW/p1732159306203639

PR #198, opened 2024-11-22
OPRUN-3605: UPSTREAM: <carry>: namespace: use privileged PSA for audit and warn levels

joelanford:
It seems we need to set audit and warn levels to privileged to avoid tripping PSA PodSecurityViolation alerts.


LalatenduMohanty (approved):
/lgtm

PR #201, opened 2024-11-26
OCPBUGS-36364: [release-4.17] Add support for using hardcoded global pull secret for bundle images

ankitathomas:
add support for using hardcoded global pull secret (openshift-config/pull-secret) for bundle images


cmd/manager/main.go
```diff
@@ -115,16 +120,41 @@ func main() {
...
-		setupLog.Error(err, "unable to create dependent label selector for cache")
-		os.Exit(1)
+	cacheOptions := crcache.Options{
+		ByObject: map[client.Object]crcache.ByObject{
+			&ocv1alpha1.ClusterExtension{}: {Label: k8slabels.Everything()},
+			&catalogd.ClusterCatalog{}:     {Label: k8slabels.Everything()},
+		},
+		DefaultNamespaces: map[string]crcache.Config{
+			systemNamespace: {LabelSelector: k8slabels.Everything()},
+		},
```
joelanford, line 144: `main` also has `DefaultLabelSelector: k8slabels.Nothing(),` in the cache options. IIRC, that was an important detail.

LalatenduMohanty: @ankitathomas looks like this needs to fixed?

ankitathomas: I've already pushed the fix on [line 145](https://github.com/openshift/operator-framework-operator-controller/pull/201/commits/8c0e2123a53ec8090d2c9e3fe1b28d242ff0c4c2#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633R145)

Besides the DefaultLabelSelector @joelanford pointed out, this is exactly the same as the [cacheOptions](https://github.com/openshift/operator-framework-operator-controller/pull/201/commits/8c0e2123a53ec8090d2c9e3fe1b28d242ff0c4c2#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633L136-L145) in use prior to this PR

ankitathomas:
> Looks good for the most part 🎉
> 
> @ankitathomas only major confusion I have is the go vendor update. Looks like there's a minor version bump for the dependencies containerregistry and ginkgo. Could you elaborate on why they're necessary?

Those dependency bumps aren't strictly necessary, but they keep our dependencies closer to their latest releases. I don't see a reason to remove them since they don't cause any issues with operator-controller performance on 4.17, so I left them in.

LalatenduMohanty (approved):
/lgtm

anik120 (approved):
Thanks for addressing the comments @ankitathomas 

/approved 

(I don't think I have approval permission for this repo but let's see...)

tmshort (commented):
/approve

PR #219, opened 2025-01-09
WIP: Fix mod vendor issue

bentito:
Seeing if this fixes at least problem with make verify throwing error on stretchr


bentito:
tide fail means verify will never run, closing

PR #223, opened 2025-01-13
OCPBUGS-47889: CVE-2025-21613 ose-olm-catalogd-container: argument injection via the URL field [openshift-4.14.z]

camilamacedo86:


camilamacedo86:
we will not move forward with this one since is tech-preview

PR #232, opened 2025-01-21
NO-ISSUE: Fix make verify and move operator-controller files

camilamacedo86:
- Fix sed to command to work on mac os
- Move openshift files to openshift/operator-controller to allow us to add the catalogd's openshift manifest in a follow up


openshift/dev_notes.md
```diff
@@ -1,4 +1,12 @@
...
-[https://github.com/openshift/operator-framework-tooling/blob/main/README.md](https://github.com/openshift/operator-framework-tooling/blob/main/README.md)
\ No newline at end of file
+[https://github.com/openshift/operator-framework-tooling/blob/main/README.md](https://github.com/openshift/operator-framework-tooling/blob/main/README.md)
+
+### (For Mac Os)
+
+To run the commands locally ensure that you have installed:
+
+```shell
+brew install findutils gnu-sed yq
```
tmshort, line 11: You actually don't need gnu-sed (i.e. `gsed`), the built in macOS `sed` works fine with the change you make.

tmshort (changes requested):

openshift/generate-manifests.sh
```diff
@@ -54,7 +54,13 @@ mkdir -p "${TMP_ROOT}/openshift"
...
 
 # Override OPENSHIFT-NAMESPACE to ${NAMESPACE}
-find "${TMP_ROOT}" -name "*.yaml" -exec sed -i "s/OPENSHIFT-NAMESPACE/${NAMESPACE}/g" {} \;
+if [[ "$(uname)" == "Darwin" ]]; then
+  # macOS
+  find "${TMP_ROOT}" -name "*.yaml" -exec sed -i '' "s/OPENSHIFT-NAMESPACE/${NAMESPACE}/g" {} +
+else
+  # Linux
+  find "${TMP_ROOT}" -name "*.yaml" -exec sed -i "s/OPENSHIFT-NAMESPACE/${NAMESPACE}/g" {} +
+fi
```
joelanford, line 63: Does the `macOS` variant of this work on linux hosts? (I think the answer is yes, but I don't know for sure).

If so, let's just change to a portable command that works on both platforms?

camilamacedo86: We tried that and failed.
So, now I pushed the condition.

tmshort: I tested it on my macOS vs Linux, and it seemed to work? I swear we had another case like this... 
We could also add `-e` before the script.
I swear we've fixed this before without resorting to checking for `Darwin`...

camilamacedo86: I tried checking older versions, and the same issue was present (likely related to the latest change in this line).
I also attempted using awk, but I wasn’t successful.

At this point, I think it might be reasonable to consider sticking with +1 for +e. I’m not sure how much effort we want to invest in removing the condition entirely.

tmshort: Maybe: https://github.com/openshift/release/blob/9712fce754de5b64f9287e3a5aea081ab29fcfde/tools/hack/golang/hack/lib/util/misc.sh#L121-L129

tmshort:
```
function os::util::sed() {
	local sudo="${USE_SUDO:+sudo}"
	if LANG=C sed --help 2>&1 | grep -q "GNU sed"; then
		${sudo} sed -i'' "$@"
	else
		${sudo} sed -i '' "$@"
	fi
}
readonly -f os::util::sed

Comment (joelanford) [reply openshift/generate-manifests.sh:63]: Let me play with it for a sec.

**Comment by tmshort** [comment]
Note that if this merges, we'll need to re-generate #224, so I'd prefer to have us accept this and then pull it into #224.
/hold

File: openshift/generate-manifests.sh (hunk 2 above)
Comment (joelanford) [reply openshift/generate-manifests.sh:63]: This worked for me with both macOS-native sed and gnu-sed.
```suggestion
find "${TMP_ROOT}" -name "*.yaml" -exec sed -i'.bak' "s/OPENSHIFT-NAMESPACE/${NAMESPACE}/g" {} \;
```

camilamacedo86:
Hi @tmshort 

We do that as follow up.
But I also would like to use this one to check the CI.

openshift/generate-manifests.sh
camilamacedo86: Worked. Cool.

tmshort: But that leaves us with a `.bak` file which we should subseuqently `.gitignore` if not already there.

camilamacedo86:
/hold cancel 

@tmshort we can merge this one now

openshift/generate-manifests.sh
joelanford: Or, even better: append an `&& find . '*.bak' -exec rm {} \;`

openshift/operator-controller/generate-manifests.sh
joelanford, line 1: I don't know that I love moving `generate-manifests.sh` into the operator-controller directory because that implies we'll have a separate `generate-manifests.sh` file for catalogd.

If the intent is that we're just taking incremental steps and that we'll eventually also consolidate the manifest generation (while still retaining the ability to generate separate sets for catalogd and operator-controller), I'm all good. I just want to make sure we don't end up in a place where we're maintaining two nearly identical scripts.

openshift/operator-controller/generate-manifests.sh
```diff
@@ -50,19 +50,20 @@ trap 'rm -rf $TMP_ROOT' EXIT
...
 cp -a "${REPO_ROOT}/config" "${TMP_ROOT}/config"
-mkdir -p "${TMP_ROOT}/openshift"
-cp -a "${REPO_ROOT}/openshift/kustomize" "${TMP_ROOT}/openshift/kustomize"
+mkdir -p "${TMP_ROOT}/openshift/operator-controller/"
+cp -a "${REPO_ROOT}/openshift/operator-controller/kustomize" "${TMP_ROOT}/openshift/operator-controller/kustomize"
 
 # Override OPENSHIFT-NAMESPACE to ${NAMESPACE}
-find "${TMP_ROOT}" -name "*.yaml" -exec sed -i "s/OPENSHIFT-NAMESPACE/${NAMESPACE}/g" {} \;
+find "${TMP_ROOT}" -name "*.yaml" -exec sed -i'.bak' "s/OPENSHIFT-NAMESPACE/${NAMESPACE}/g" {} \;
+
```
joelanford, line 58: As Todd mentioned in the other thread, I think we need to deal with the backup files that will be created after the `find ... sed` command runs.

Maybe add another command like this:
```
find "${TMP_ROOT}" -name "*.bak" -exec rm {} \;
```

camilamacedo86: Let's see if it will pass the tests first, and then we will do that.

openshift/operator-controller/generate-manifests.sh
camilamacedo86: I believe the process is already challenging enough. The tests take too long to run, and there is too much flakiness. It becomes difficult to determine whether a failure is due to a genuine error or just a flake.

Additionally, I’ve already tried copying catalogd and creating its own generate-manifest.sh, but it’s still not working. There’s some detail missing, and I haven’t been able to pinpoint it.

So the idea is to go with baby steps and make things work. First, we ensure everything works. Once that’s in place, we can focus on refining and optimizing the process further. WDYT? Does that make sense?

openshift/operator-controller/generate-manifests.sh
tmshort: Or add it to `.gitignore`

openshift/operator-controller.Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN make go-build-local
 FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
 USER 1001
 COPY --from=builder /build/bin/operator-controller /operator-controller
-COPY openshift/manifests /openshift/manifests
+COPY openshift/operator-controller/manifests /openshift/operator-controller/manifests
```
tmshort, line 9: The manifest destination should remain at `/openshift/manifests`. Only the source _should_ change.

openshift/registry.Dockerfile
```diff
@@ -10,7 +10,7 @@ FROM registry.ci.openshift.org/ocp/4.18:base-rhel9
 USER 1001
 COPY --from=builder /build/registry /registry
 COPY --from=builder /build/push /push
-COPY openshift/manifests /openshift/manifests
+COPY openshift/operator-controller/manifests /openshift/operator-controller/manifests
```
tmshort, line 13: Again, the manifest destination should remain at `/openshift/manifests`. Only the source should change.

camilamacedo86: I see good point !!!

openshift/operator-controller/generate-manifests.sh
camilamacedo86: @tmshort @joelanford 

I added a line to remove the .bak files, but since they’re created in TMP_ROOT, they’ll be cleaned up anyway when the trap deletes TMP_ROOT. So technically, the extra remove line isn’t really needed, but it’s there now, so it doesn’t hurt. 😊

openshift/operator-controller.Dockerfile
camilamacedo86: Done Good catcher

openshift/operator-controller/generate-manifests.sh
tmshort: Yeah, if it's done in TMP_ROOT... so not necessary to clean up...

openshift/generate-manifests.sh
```diff
@@ -54,7 +54,8 @@ mkdir -p "${TMP_ROOT}/openshift"
 cp -a "${REPO_ROOT}/openshift/kustomize" "${TMP_ROOT}/openshift/kustomize"
 
 # Override OPENSHIFT-NAMESPACE to ${NAMESPACE}
-find "${TMP_ROOT}" -name "*.yaml" -exec sed -i "s/OPENSHIFT-NAMESPACE/${NAMESPACE}/g" {} \;
+find "${TMP_ROOT}" -name "*.yaml" -exec sed -i'.bak' "s/OPENSHIFT-NAMESPACE/${NAMESPACE}/g" {} \;
+find "${TMP_ROOT}" -name "*.bak" -exec rm {} \;
```
tmshort, line 58: actually... no exec is needed, the action could just be `-delete`:
```
find "${TMP_ROOT}" -name "*.bak" -delete

Comment (camilamacedo86) [reply openshift/generate-manifests.sh:58]: https://github.com/openshift/operator-framework-operator-controller/pull/232#discussion_r1925491963

If that passes in the CI, we remove it in the follow-up.
I have the commit with the catalogd already, but I cannot push another PR with the same commits, etc, required so if you agree, I think that is the easy away.

Comment (tmshort) [reply openshift/generate-manifests.sh:58]: Not blocking... just an optimization

File: openshift/Makefile [hunk 7]
```diff
@@ -15,7 +15,7 @@ verify: ## Run downstream-specific verify
 
 .PHONY: manifests
 manifests: $(KUSTOMIZE) $(YQ)
-	$(DIR)/generate-manifests.sh
+	$(DIR)/operator-controller/generate-manifests.sh
```
LalatenduMohanty, line 18: So instead of /operator-framework-operator-controller/blob/main/openshift/generate-manifests.sh we are targeting /operator-framework-operator-controller/blob/main/openshift/operator-controller/generate-manifests.sh

openshift/operator-controller/generate-manifests.sh
LalatenduMohanty: @camilamacedo86 What about moving catalogd/openshift directory to this repo? Are you planning to do that in separate commit or PR? https://github.com/openshift/operator-framework-catalogd/tree/main/openshift

LalatenduMohanty:
@camilamacedo86 I would suggest that we break this PR in to two PRs one for `Fix sed to command to work on mac os` and another `Move openshift files to openshift/operator-controller to allow us to add the catalogd's openshift manifest in a follow up`

tmshort (approved):
/lgtm
As a small step

LalatenduMohanty (approved):
/lgtm

tmshort:
/title NO-ISSUE: Fix make verify and move operator-controller files

tmshort:
/retitle NO-ISSUE: Fix make verify and move operator-controller files

PR #233, opened 2025-01-22
NO-ISSUE: Add catalogd manifests to start to run those tests

camilamacedo86:
- Move to openshift/catalogd the specific manifest under: https://github.com/openshift/operator-framework-catalogd/tree/main/openshift 
- Add call to generate catalogd manifest to 'make manifest'. Make verify test is now done for catalogd and operator-controller Openshift's manifests

### DIFF
--- a/openshift/Makefile
+++ b/openshift/Makefile
@@ -16,6 +16,7 @@ verify: ## Run downstream-specific verify
 .PHONY: manifests
 manifests: $(KUSTOMIZE) $(YQ)
 	$(DIR)/operator-controller/generate-manifests.sh
+	$(DIR)/catalogd/generate-manifests.sh
 
 .PHONY: verify-manifests
 verify-manifests: manifests
@@ -35,3 +36,8 @@ test-e2e: ## Run the e2e tests.
 	$(DIR)/operator-controller/build-test-registry.sh $(E2E_REGISTRY_NAMESPACE) $(E2E_REGISTRY_NAME) $(E2E_REGISTRY_IMAGE)
 	cd $(DIR)/../; \
 	go test $(DOWNSTREAM_E2E_FLAGS) ./test/e2e/...;
+	# Run e2e tests for catalogd
+	# TODO: Add build of the testdata/test-catalog.Dockerfile to openshift/release
+	# See that catalogd e2e tests uses testdata/test-catalog.Dockerfile
+	# More info: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/operator-framework-catalogd/openshift-operator-framework-catalogd-main.yaml#L26-L29
+	#$(MAKE) test-e2e -C $(DIR)/../catalogd/
--- a/openshift/catalogd.Dockerfile
+++ b/openshift/catalogd.Dockerfile
@@ -1,0 +1,12 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19 AS builder
+WORKDIR /build
+COPY . .
+RUN make go-build-local
+
+FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
+USER 1001
+COPY --from=builder /build/bin/catalogd /catalogd
+COPY openshift/catalogd/manifests /openshift/manifests
+
+LABEL io.k8s.display-name="OpenShift Operator Lifecycle Manager Catalog Controller" \
+      io.k8s.description="This is a component of OpenShift Container Platform that provides operator catalog support."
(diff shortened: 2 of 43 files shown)


openshift/catalogd.Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.23-openshift-4.19 AS builder
```
LalatenduMohanty, line 1: I guess this is 1.23.2 version of the go.

openshift/catalogd.Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+WORKDIR /build
+COPY . .
+RUN make go-build-local
+
+FROM registry.ci.openshift.org/ocp/4.19:base-rhel9
+USER 1001
+COPY --from=builder /build/bin/catalogd /catalogd
+COPY openshift/catalogd/manifests /openshift/manifests
```
LalatenduMohanty, line 9: :+1:

openshift/catalogd/generate-manifests.sh
```diff
@@ -0,0 +1,132 @@
...
+    # Some file names (namely CRDs) don't end in .yml - make them
+    if ! [[ "$new_name" =~ yml$ ]]; then
+      new_name="${new_name}".yml
+    fi
+    if [[ "$f" != "$new_name" ]]; then
+      # Rename
+      mv "$f" "${new_name}"
+    fi
+  done
+)
```
LalatenduMohanty, line 132: I am assuming that it is a copy from https://github.com/openshift/operator-framework-catalogd/blob/main/openshift/generate-manifests.sh. So not going through the details.

openshift/catalogd.Dockerfile
camilamacedo86: This file is the same merged at https://github.com/openshift/operator-framework-catalogd and passing in the CI/prow jobs there. See; https://github.com/openshift/operator-framework-catalogd/blob/main/openshift/Dockerfile

openshift/catalogd/generate-manifests.sh
camilamacedo86: Yes :-) 

 it was copied from there and adjusted accordingly. The same changes we applied to the operator-controller manifests were made here as well, since the OpenShift manifests are now located under openshift/catalogd, and the catalogd project manifests are in the ROOT/catalogd directory instead of ROOT.

mainly the same changes done for operator-controller here: https://github.com/openshift/operator-framework-operator-controller/pull/232/commits/131d4c4057511ed000dbf65a35b034feb375d0b2 are applied for Openshift manifest of catalogd.

testdata/catalogs/test-catalog.Dockerfile
joelanford, line 1: I may be mistaken, but I think all of these testdata files are already present in `/catalogd/testdata`.

camilamacedo86: It seems caused by the script, in the prepared step where we used `git mv`

The git mv did not move testdata dir from catalogd to operator-controller because operator-controller have testdata directory already.

camilamacedo86: @tmshort ^ I think this one is important for you to be aware of.

The next step here is to investigate if we really need this `test-catalog`. Specifically:
- Should the tests in upstream for `catalogd` be using it?  
- OR is it not required, and the `testdata` mock in upstream doesn't need to include it here?

This distinction will help clarify whether it’s something specific to downstream or necessary across both.

openshift/Makefile
```diff
@@ -35,3 +36,8 @@ test-e2e: ## Run the e2e tests.
 	$(DIR)/operator-controller/build-test-registry.sh $(E2E_REGISTRY_NAMESPACE) $(E2E_REGISTRY_NAME) $(E2E_REGISTRY_IMAGE)
 	cd $(DIR)/../; \
 	go test $(DOWNSTREAM_E2E_FLAGS) ./test/e2e/...;
+	# Run e2e tests for catalogd
+	# TODO: Add build of the testdata/test-catalog.Dockerfile to openshift/release
+	# See that catalogd e2e tests uses testdata/test-catalog.Dockerfile
+	# More info: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/operator-framework-catalogd/openshift-operator-framework-catalogd-main.yaml#L26-L29
+	#$(MAKE) test-e2e -C $(DIR)/../catalogd/
```
camilamacedo86, line 43: We will need to change that to use the same from catalogd/testdata/test-catalog.Dockerfile instead

LalatenduMohanty (approved):
/lgtm

PR #235, opened 2025-01-23
NO-ISSUE: Upgrade OCP images from 4.18 to 4.19

camilamacedo86:


camilamacedo86:
@dtfranz ^

tmshort (approved):
This one is probably not so critical to be 4.19, but we should be consistent.
/lgtm

PR #237, opened 2025-01-24
OCPBUGS-48830: resolve issue with pre-mature mounting of trusted CA configmap

joelanford:


tmshort:
Did you also want to do with with the service-ca.crt?

tmshort:
And we likely want to do it for catalogd.

tmshort:
> And we likely want to do it for catalogd.

Already done: https://github.com/openshift/operator-framework-catalogd/pull/120

PR #244, opened 2025-01-28
UPSTREAM: <carry>: refactor how CAs work to accommodate OCP nuances

joelanford:
Inventory of changes:
- Only mount and use service-cas for use with operator-controller's http client used to connect to catalogd
- Remove trusted-cas ConfigMap from catalogd and operator-controller, replace with mounted /etc/pki directory (thus putting trusted CAs in our pod's system root store)
- Make all cert-related mounts read-only.
- In operator-controller, only use the `--ca-cert-dir` flag to verify catalogd's web server.

I expect this to initially fail catalogd's e2e tests because the registry included in the e2e test uses service-ca-signed certs, which are not trusted by the containers/image library as a result of the changes in this PR. My intention with that problem is to update catalogd's e2e to push/pull using OCP's existing image registry.

I also expect to update this once we validate that it works correctly to solve some open issues around:
- CA trust for catalogd communication from operator-controller (We should now be waiting for the service-ca.crt key to exist before our pod starts. I'm not sure this was always happening before)
- CA trust (including additional CAs configured in images.config.openshift.io and proxies.config.openshift.io objects, by just fully mounting the node's /etc/pki directory)

Once we validate that this sort of mount setup works, we will need to go back upstream and stop using `caCertsDir` in the `containers/image` SystemContext. Otherwise, I'm hopeful the rest of the changes in this PR will work and can be merged.


cmd/operator-controller/main.go
```diff
@@ -105,7 +105,7 @@ func main() {
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "", "The address for the metrics endpoint. Requires tls-cert and tls-key. (Default: ':8443')")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.StringVar(&caCertDir, "ca-certs-dir", "", "The directory of TLS certificate to use for verifying HTTPS connections to the Catalogd and docker-registry web servers.")
+	flag.StringVar(&caCertDir, "ca-certs-dir", "", "The directory of TLS certificate to use for verifying HTTPS connections to the catalogd")
```
camilamacedo86, line 108: @joelanford should we change it?
I think we need to do this change upstream; otherwise, we will face conflicts.

PR #252, opened 2025-01-30
UPSTREAM: <carry>: OCPBUGS-49410: Enable OCP metrics collection by default

azych:
Enables OCP to collect Prometheus metrics for both `catalogd` and `operator-controller` by default. This is accomplished via ServiceMonitor CRs which are now created for both projects.

Tested on both 4.19 and 4.19-nightly (aka main) versions:

Screenshot 2025-02-03 at 09 57 23
Screenshot 2025-02-03 at 09 57 34
Screenshot 2025-02-03 at 09 57 44

Fixes OCPBUGS-49410


openshift/catalogd/manifests/20-role-openshift-catalogd-metrics.yml
```diff
@@ -0,0 +1,20 @@
+---
```
camilamacedo86, line 1: Very cool 👍 
we will need to have the same for operator-controller as well 👍

camilamacedo86: Another nit:
For this one pass in the CI we need to add a commit message like 
`UPSTREAM: <carry>: MESSAGE`

That is because of the bumper
How we sync the things from upstream to downstream
And to allow the tool/bumper to be aware that this needs to be carried on.

joelanford (changes requested):
Two things:
1. We should do a similar thing for operator-controller
2. We'll need these included in the kustomize directory somewhere. The manifests directories are generated from kustomize.

openshift/catalogd/manifests/20-role-openshift-catalogd-metrics.yml
azych: hmm, so no OCPBUGS-49410 at all? or should it be included in the message?
or for this we just use UPSTREAM carry and then for 4.19 for example we can do OCPBUGS-49410?

azych:
> Two things:
> 
>     1. We should do a similar thing for operator-controller
> 
yes, I intend to, it's a Draft-WIP at this point until I can verify that everything is working and come up with manifests for operator-controller

>     2. We'll need these included in the kustomize directory somewhere. The manifests directories are generated from kustomize.
good point, I'll do that before removing Draft-WIP status of this PR

openshift/catalogd/manifests/22-servicemonitor-catalogd.yml
```diff
@@ -0,0 +1,28 @@
...
+  endpoints:
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      interval: 30s
+      path: /metrics
+      port: metrics
+      scheme: https
+      tlsConfig:
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+        certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
+        keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
```
tmshort, line 21: My gut tells me these aren't necessary... I could be wrong.

tmshort:
Fundamentally, these files are supposed to be created through the kustomize mechanism, which means adding resources to the `openshift/catalogd/kustomize/overlays/openshift` directory and appropriate kustomization.yaml updates and patchtes.

openshift/catalogd/manifests/22-servicemonitor-catalogd.yml
camilamacedo86: Because all inside the cluster (Prometheus and OLM compontets) probably do not need to pass the certs.
That is good point. It might work with RBAC permissions only.

openshift/catalogd/manifests/20-role-openshift-catalogd-metrics.yml
camilamacedo86: 4.19, 4.20 will be automatic synced 
Then, when you cherry-pick for 4.18 you would do like `UPSTREAM: PRID: Message` you will need to change that.

openshift/catalogd/manifests/20-role-openshift-catalogd-metrics.yml
```diff
@@ -0,0 +1,20 @@
...
+  - apiGroups:
+      - ""
+    resources:
+      - services
+      - endpoints
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
```
joelanford, line 20: We may also need to provide RBAC for the `/metrics` endpoints as a non resource URL.

openshift/catalogd/manifests/22-servicemonitor-catalogd.yml
joelanford: These would be the client certificates that prometheus presents to our endpoint when it scrapes it. I don't think we have configured client-cert-based auth. I would think the service account token specified in line 13 would be what we use to authenticate.

azych:
> Fundamentally, these files are supposed to be created through the kustomize mechanism, which means adding resources to the `openshift/catalogd/kustomize/overlays/openshift` directory and appropriate kustomization.yaml updates and patches.

[good point, I'll do that before removing Draft-WIP status of this PR](https://github.com/openshift/operator-framework-operator-controller/pull/252#issuecomment-2624704061)

openshift/catalogd/manifests/20-role-openshift-catalogd-metrics.yml
azych: IIRC that's already being provided by `prometheus-k8s` clusterrole

azych:
@camilamacedo86 @joelanford @tmshort 

To clarify & try to find the way forward:
* this general setup and tlsconfig has been verified prior to issuing this Draft PR on a live 4.19 cluster to work correctly
* if the setup changed and we're using self-signed cert in 'main' (probably a bug in itself), then should we want to go ahead with the changes described here just for 4.19? do we want to do it in this PR?
* what about https://github.com/openshift/operator-framework-operator-controller/pull/252#discussion_r1935847984 ? Is there a way not to have the changes done in main automatically synced down?
* If we want and need to solve this in 'main' first for some reason, this means dependency on https://github.com/openshift/operator-framework-operator-controller/pull/255 potentially, true?

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - apiGroups:
+      - ""
+    resources:
+      - services
+      - endpoints
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
```
camilamacedo86, line 19: We should not scaffold the rules. The roles are scaffolded already. 
See: https://github.com/operator-framework/operator-controller/blob/e77c53c92d8f28a6c9419739b6651ffc0f79e5c9/test/e2e/metrics_test.go#L54-L59

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_servicemonitor.yaml
```diff
@@ -0,0 +1,28 @@
...
+        # caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+        # certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
+        # keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
+        # serverName: catalogd-service.openshift-catalogd.svc
+  namespaceSelector:
+    matchNames:
+      - openshift-catalogd
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: catalogd
```
camilamacedo86, line 28: I think we should do like

```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  labels:
    control-plane: operator-controller-controller-manager
  name: controller-manager-metrics-monitor
  namespace: system
spec:
  endpoints:
    - path: /metrics
      port: https
      scheme: https
      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
      tlsConfig:
        caFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt  # Uses Kubernetes CA for TLS verification
        serverName: operator-controller-service.<namespace>.svc.cluster.local
  selector:
    matchLabels:
      control-plane: operator-controller-controller-manager
```

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role_binding.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: metrics-rolebinding
+  namespace: openshift-catalogd
+  labels:
+    app.kubernetes.io/name: catalogd
+    app.kubernetes.io/part-of: olm
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: metrics-role
```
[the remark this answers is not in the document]
azych: that could be the case if we had an external/user tool wanting to read metrics.

We have the openshift monitoring machinery already set up in cluster and we need to be binding this role with existing `prometheus-k8s` serviceaccount like we are doing here:

```yaml
subjects:
  - kind: ServiceAccount
    name: prometheus-k8s
    namespace: openshift-monitoring
```

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
azych: are you talking about `operator-controller-metrics-reader`, because it doesn't really do it:
```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: operator-controller-metrics-reader
rules:
  - nonResourceURLs:
      - /metrics
    verbs:
      - get
```

If not, please link to manifest that you mean that already defines those permissions in a role in our namespaces we use for catalogd and operator-controller

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_servicemonitor.yaml
azych: IMHO we should be using the same prometheus `service-ca` for validation that all other cluster service monitors seem to use, unless we cannot fix the current self-signed cert situation in a reasonable time (AFAIK PR with fix is already created)

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
camilamacedo86: But why do we need to give those permissions?
Prometheus needs access to the metrics endpoint and that is it.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_servicemonitor.yaml
camilamacedo86: Ignore I was silly we need to pass the certs or skip true

openshift/catalogd/manifests/11-rolebinding-openshift-config-catalogd-manager-rolebinding.yml
```diff
@@ -13,5 +13,5 @@ roleRef:
   name: catalogd-manager-role
 subjects:
   - kind: ServiceAccount
-    name: catalogd-controller-manager
```
camilamacedo86, line 16: Why are we changing the name of the SA?

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
azych: it does, but we're configuring a service monitor (as opposed to pod monitor for example), which from what I understand needs to:
- discover services that match the `serviceMonitorSelector` in the configured namespace (I actually added a comment to your doc PR about it a couple of days ago: https://github.com/operator-framework/operator-controller/pull/1524#issuecomment-2622593325)
- discover the matching endpoint(s) configured for the specified port on the discovered service(s)
- scrape the endpoint(s) and likely do more that requires pod permissions

Practically speaking, without those permissions, we'd be getting error messages in our prometheus-k8s instances logs and no scraping will happen

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_servicemonitor.yaml
azych: I think if we don't skip true, we'd need to at least provide a valid CA for validation

openshift/catalogd/manifests/11-rolebinding-openshift-config-catalogd-manager-rolebinding.yml
azych: good catch. If you look at PR changes, I'm not changing anything within config folder so it's interesting. I'll look into it

openshift/catalogd/manifests/20-role-openshift-catalogd-metrics.yml
azych: I'm leaving the title as-is, 4.18 changes would require additional work in order to work and therefore should be synced automatically. If (given internal discussions) we want to make those changes for 4.18, it'd need to happen in a separate PR.

azych:
@joelanford @tmshort or any other reviewer:  I'd appreciate some guidance on proper labeling and jira ticket changes (eg. target version)

openshift/operator-controller/manifests/14-rolebinding-openshift-config-operator-controller-manager-rolebinding.yml
```diff
@@ -13,5 +13,5 @@ roleRef:
   name: operator-controller-manager-role
 subjects:
   - kind: ServiceAccount
-    name: operator-controller-controller-manager
+    name: controller-manager
```
camilamacedo86, line 16: We should not change the name, right?

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
camilamacedo86: The problem here shows that we are not using the ` operator-controller-metrics-reader.`
Prometheus only need to have access to the metrics endpoint as far I understand, not the pods or services.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/patches/manager_namespace_monitored.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: system
+  labels:
+    openshift.io/cluster-monitoring: 'true'
```
camilamacedo86, line 6: Can you please add a space at the end of the file here?

openshift/catalogd/kustomize/overlays/openshift/openshift-config/rbac/catalogd_manager_role_binding.yaml
```diff
@@ -11,5 +11,5 @@ roleRef:
   name: manager-role
 subjects:
 - kind: ServiceAccount
-  name: controller-manager
+  name: catalogd-controller-manager
```
camilamacedo86, line 14: Why are we changing the name here?

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
azych: @camilamacedo86 

Most of my arguments as to why we need the approach the PR is currently using has already been mentioned here: https://github.com/openshift/operator-framework-operator-controller/pull/252#discussion_r1937124453

The main disconnect here seems to be that we're not using plain Prometheus running in our namespace, but leveraging OCP setup that's already running in cluster, which uses prometheus-operator and instruments scraping through `ServiceMonitors`.

`operator-controller-metrics-reader` duplicates what `prometheus-k8s` (OCP prometheus) has included in its `prometheus-k8s` clusterrole, so all that we need to provide additionally is a local (namespace) role and binding for it to have the permissions for services, endpoints and pods that it needs additionally.

From I understand with  `operator-controller-metrics-reader` we're trying to accomplish something different here - having a role that's ready for users that want to setup their own monitoring to the /metrics endpoint. Extending it with those additional permissions that a `ServiceMonitor` need might not make sense because not every user would want to actually use a `ServiceMonitor` and as you say - plain Prometheus should only be needing `/metrics` access.

With the new role, we're just adding scaffolding necessary for downstream setup to have those metrics scraped and because this setup is downstream-only, I wouldn't want to make those permissions/role changes upstream.

openshift/operator-controller/manifests/14-rolebinding-openshift-config-operator-controller-manager-rolebinding.yml
azych: must've been tangled during a rebase, good spot

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_servicemonitor.yaml
camilamacedo86: You requested that my initial comment not be marked as out of context, so I am leaving it as is. However, I am unclear as to why this thread has not been marked as resolved, given that similar comments have been. Could you clarify what additional concerns remain?

**To reiterate, my initial comment was a mistake, which I acknowledged both here and via Slack shortly after. I also understand that you have not acknowledged this yet. So, again, it can be disregarded entirely.**

Screenshot 2025-02-03 at 15 34 12

That said, I would appreciate understanding how this benefits the review process and others. Additionally, could you clarify why other review threads were marked as resolved while this one remains open? I suggest marking resolved discussions accordingly and keeping only those requiring attention open.

camilamacedo86: You requested that my initial comment not be marked as out of context, so I am leaving it as is.

**To reiterate, my initial comment was a mistake, which I acknowledged both here and via Slack shortly after. I also understand that you have acknowledged this already. So if not, to make it clear, it can be disregarded entirely.**

Screenshot 2025-02-03 at 15 34 12

Additionally, I noticed that other threads have been marked as resolved, while this one remains open. I do not understand the difference—**what else needs to be addressed here?** Is there anything else that we need to discuss about this topic?  

I suggest maintaining consistency by marking resolved discussions accordingly and keeping only those that still require attention open. This would help others review your PR more effectively by focusing on the discussions that still need attention.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
camilamacedo86: That is the result we have both downstream:

- metrics-role
- metrics-reader

And now the same for catalogd 😞 (see: [auth_proxy_client_clusterrole.yaml](https://github.com/operator-framework/operator-controller/blob/main/catalogd/config/base/rbac/auth_proxy_client_clusterrole.yaml)).

**Could this lead to confusion, redundancy, or unnecessary maintenance?**

IMHO, we could:

- **Option 1:** Patch it here to add the verbs to the existing role.
OR
- **Option 2:** Add the verbs to the upstream role.

Anyway, I’ll leave it to @joelanford and @tmshort to decide.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- catalogd_metrics_role.yaml
+- catalogd_metrics_role_binding.yaml
+- catalogd_servicemonitor.yaml
```
camilamacedo86, line 4: NIT: Can you please add a space at the end of the file to remove the symbol?

openshift/catalogd/manifests/11-rolebinding-openshift-config-catalogd-manager-rolebinding.yml
camilamacedo86: Can you please reopen this one? 
It still changed. Did you check why?

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
azych: @camilamacedo86 can you share why you don't think my arguments against both the options that you specify here are not valid to you? I'm not saying they aren't, but you're offering options despite my arguments explaining why I don't think it makes sense to do what you're proposing so it would be good to understand why you think that and if I'm wrong somehow.

All of my arguments are in this thread btw.

camilamacedo86: Hi @azych 

I reread your argument, and from one perspective, I see your point—it’s a valid POV. However, could we at least consider a name that makes its purpose more intuitive? For example, something like ocp-cluster-monitoring-role instead of just metrics-role would provide more clarity.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: metrics-monitor-role
```
camilamacedo86, line 4: Could we use a more intuitive name, like ocp-cluster-monitoring-role?

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role_binding.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: metrics-monitor-rolebinding
```
camilamacedo86, line 4: Could we use a more intuitive name, like ocp-metrics-monitoring-rolebinding?

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_role.yaml
camilamacedo86: I left a comment/review for the specific request to change the name for something more intuitive that brings claroty. I think you’ve convinced me, so feel free to mark it as resolved. This won’t be a blocker for me to give my LGTM here.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/kustomization.yaml
```diff
@@ -7,10 +7,12 @@ resources:
 - ../../../../../../config/base/crd
 - ../../../../../../config/base/rbac
 - ../../../../../../config/base/manager
+- metrics
```
camilamacedo86, line 10: Should the metrics be under namespace config?
Would it not be better if we have a dir for metrics at the same level as the others?

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role.yaml
azych: We're currently not adding `ocp-` prefix to any objects that we create (even the ones exclusive to OCP) and `-ing` doesn't bring any additional value in my mind so I'd leave it as-is unless you feel strongly about this and consider it a blocker.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
azych: because `openshift-config/rbac` stores rbac for `openshift-config` namespace and all of the `metrics` objects that we're creating are being created in our dedicated `catalogd`/`operator-controller` namespace

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/kustomization.yaml
azych: I think they should, because all those objects are being created in the `olmv1-ns`, just like all the other objects that are already placed there, see also: https://github.com/openshift/operator-framework-operator-controller/pull/252#discussion_r1944289895

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role_binding.yaml
azych: already discussed here: https://github.com/openshift/operator-framework-operator-controller/pull/252#discussion_r1944285206

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role.yaml
camilamacedo86: We don’t have many things that are specific to OCP, but this would be the first case where we introduce a potential duplication.

If we merge this as-is, we will end up with:

- `metrics-reader`
- `metrics-monitor-role`

#### Concerns:
- Once merged and released, we won’t be able to change the name.
- Naming a role with the word **"role"** isn’t ideal—we generally avoid that for clarity.
- The distinction between these two roles isn’t clear, making it confusing why both are needed for metrics.
- Since this is specific to **OpenShift Monitoring**, we should choose a name that:
  - Clearly reflects its purpose.
  - Avoids redundancy by omitting `"role"` from the name.

I’d suggest renaming it to something more descriptive that better aligns with its function in OpenShift Monitoring.

camilamacedo86: We don’t have many things that are specific to OCP, but this might be the first case where we introduce some duplication.

If we merge this as-is, we will have:

- `metrics-reader`
- `metrics-monitor-role`

#### Some concerns:
- Once merged and released, renaming will be difficult.
- Including `"role"` in the name might not be ideal—IMHO, we might want to avoid that for clarity.
- The difference between these two roles isn’t immediately clear, which might make it confusing to understand why both are needed for metrics.
- Since this is specific to **OpenShift Monitoring**, maybe we could choose a name that:
  - Clearly reflects its purpose.
  - Avoids redundancy by not including `"role"` in the name.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role.yaml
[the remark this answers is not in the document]
camilamacedo86: We need to clarify how we are organizing these directories.

Screenshot 2025-02-06 at 08 42 01

### **Questions to Consider:**
- Does each directory represent a **feature/customization**?
  - We have **specific customizations** for `opc ns`.
  - We have **specific customizations** for `config`.
  - We have **specific customizations** for `catalogs`.
- Now, we are adding **specific customizations for metrics**—should it follow the same structure?
- If so, shouldn't **metrics** be placed at the same level as the others instead of being added below everything?
- What is the **logic/purpose** behind how we are organizing these directories?

@tmshort is the person to help us out here.
But looking today I think his idea was per "kind/feature" that we customize so we might only need to put the metrics in the same level of the others.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/kustomization.yaml
camilamacedo86: We need to clarify how we are organizing these directories.

Screenshot 2025-02-06 at 08 42 01

### **Questions to Consider:**
- Does each directory represent a **feature/customization**?
  - We have **specific customizations** for `opc ns`.
  - We have **specific customizations** for `config`.
  - We have **specific customizations** for `catalogs`.
- Now, we are adding **specific customizations for metrics**—should it follow the same structure?
- If so, shouldn't **metrics** be placed at the same level as the others instead of being added below everything?
- What is the **logic/purpose** behind how we are organizing these directories?

@tmshort is the person to help us out here.
But looking today I think his idea was per "kind/feature" that we customize so we might only need to put the metrics in the same level of the others.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role.yaml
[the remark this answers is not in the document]
camilamacedo86: Looking at it again, it seems like we are organizing by **feature**. If that's the case, then the RBAC should still remain under **metrics**.  

However, we should confirm this: [discussion link](https://github.com/openshift/operator-framework-operator-controller/pull/252#discussion_r1944328477).

camilamacedo86 (commented):
### Requesting Changes  

Great work! :rock: I think we’re close to getting this merged. From my perspective, we just need to address the following two points:  

1. **Naming Consistency for Metrics Roles**  
   Since we will end up with two roles for metrics (`metrics-reader`, `metrics-monitor-role`), should we use a name that better clarifies its purpose?  
   For example, should it include **"openshift-monitoring"** or something similar to indicate its intent?  
   👉 Related discussion: [comment link](https://github.com/openshift/operator-framework-operator-controller/pull/252#discussion_r1944303590)  

2. **Directory Organization for OCP Customizations**  
   How should the customizations for OCP be organized? Should all OCP monitoring-related configurations be placed under:  
   - `openshift/catalogd/kustomize/overlays/openshift/metrics`  
   - **OR** should they be in `openshift/catalogd/kustomize/overlays/openshift/monitoring`, since we seem to be organizing by **feature** and this customization is meant to integrate with OCP monitoring?  

   @Todd would be the best person to clarify this, as he originally structured these directories and likely has a clear vision of their intended organization.  
   👉 Related discussion: [comment link](https://github.com/openshift/operator-framework-operator-controller/pull/252#discussion_r1944328477)  

### CI & Commit Requirements  

- The PR must pass CI checks:  
  - I recommend **rebasing with `main`** to include the latest changes/fixes.  
- The `ci/prow/verify-commits` check is failing due to the **commit message format**.  
  - The commit message must follow the pattern:  
    ```text
    UPSTREAM: <carry>: <commit message>
    ```
  - Example:  
    ```text
    UPSTREAM: <carry>: Enable OCP metrics collection by default
    ```  

Once these points are addressed, we should be good to go! 🚀

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role.yaml
camilamacedo86: > I am not looking to change this convention now and potentially break things tied to cluster-olm-operator.

Cluster-olm-operator is OLM V0
Nothing here should to be tied to there.

**It's a NIT**, but the problem is after will be very hard to rename. So, it would be great if we could give it a name that clarifies that it is required because we need to integrate with OCP Monitoring. That might help us and users in the future.

azych: > > I am not looking to change this convention now and potentially break things tied to cluster-olm-operator.
> 
> Cluster-olm-operator is OLM V0 Nothing here should to be tied to there.

And yet it is because cluster-olm-operator actually operates `catalogd` and `operator-controller` and is tied to some of the roles we define here to be able to bind them, you can discover it for yourself here for example: https://github.com/openshift/cluster-olm-operator/blob/main/manifests/0000_51_olm_02_operator_clusterrole.yaml 

> **It's a NIT**, but the problem is after will be very hard to rename. So, it would be great if we could give it a name that clarifies that it is required because we need to integrate with OCP Monitoring. That might help us and users in the future.

You're saying it's a nit, but you still expect a name change, which makes it a blocker

camilamacedo86: I believe [this file](https://github.com/openshift/cluster-olm-operator/blob/main/manifests/0000_51_olm_02_operator_clusterrole.yaml) might be there to support migration between versions or to ensure that only one of them runs at a time. I’m not entirely sure, though—it’s a great question to ask why there are OLMv1-related elements in OLMv0.

This is just a small nitpick, but in my opinion, we might be able to find a more suitable name. Naming is always tricky, though, and I’m happy to go with whatever works best for everyone.

Just to clarify, I’m not blocking anything here. As I’ve mentioned before, I think we need input from others regarding the overall organization and TBD items. The main requirement is that the PR needs to pass CI, which is a hard blocker. But if everyone is okay with this name, I’m happy to move forward with it too.

tmshort: @camilamacedo86, Incorrect, `cluster-olm-operator` _is_ OLMv1., but also should not be involved here.

If you look at the resulting manifests, you will see the name is updated by `kustomize` to be: `catalogd-metrics-monitor-role`, `catalogd-` is prefixed.

tmshort: See: https://github.com/openshift/operator-framework-operator-controller/pull/252/files#diff-07b1239dfc963ac0ebf32b97e7199fc87208bd7272e0296515999406ad5af73a

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_servicemonitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+  labels:
+    app.kubernetes.io/name: catalogd
+    app.kubernetes.io/part-of: olm
+    openshift.io/cluster-monitoring: 'true'
+spec:
+  endpoints:
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      interval: 30s
+      path: /metrics
+      port: metrics
```
tmshort, line 14: What is this value (numerically)? Does it line up with the catalogd service (`7443`)?

openshift/catalogd/manifests/22-servicemonitor-catalogd.yml
tmshort: Ye, that's why I think they're unnecessary, they'd be used for client authentication, which we don't have turned on, and very likely would require adding _another_ CA...

tmshort:
For the `ci/prow/verify-commits` job, update your commit headline to include the `UPSTREAM: <carry>:` tag:
```
UPSTREAM: <carry>: Enable OCP metrics collection by default
```

openshift/operator-controller/kustomize/overlays/openshift/olmv1-ns/metrics/operator_controller_metrics_monitor_servicemonitor.yaml
```diff
@@ -0,0 +1,25 @@
...
+  name: metrics-monitor
+  labels:
+    control-plane: operator-controller-controller-manager
+    openshift.io/cluster-monitoring: 'true'
+spec:
+  endpoints:
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      interval: 30s
+      path: /metrics
+      port: https
```
tmshort, line 13: What is this value numerically? Does it match with the operator-controller metrics port (8443)?

tmshort: Note that the op-con metrics port is different than catalogd.

openshift/operator-controller/kustomize/overlays/openshift/olmv1-ns/patches/manager_namespace_monitored.yaml
tmshort, line 1: This file should really be in a patch format.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/patches/manager_namespace_monitored.yaml
tmshort, line 1: This file should really be in a patch format.

tmshort:
It does appear that https://github.com/openshift/cluster-olm-operator/pull/103 will be needed to pass the e2e's. I've already reviewed it.

@camilamacedo86, I have no idea who `@todd` is, but it isn't me.
@azych:

Location: Since we only have one set of manifests, there is no need for overlays. Patches should be in the `olmv1-ns` directory as is done here in this PR.

Naming: Look at the generated manifests, `kustomize` prefixes the name of the component. The names are fine as they are, as they are prefixed with either `catalogd` or `operator-controller`. I don't think we need to make any changes 

Commit headline: it's been commented on numerous times.

openshift/operator-controller/kustomize/overlays/openshift/olmv1-ns/metrics/operator_controller_metrics_monitor_servicemonitor.yaml
azych: yes, see https://github.com/openshift/operator-framework-operator-controller/pull/252#discussion_r1944945062

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_role.yaml
camilamacedo86: >@camilamacedo86, Incorrect, cluster-olm-operator is OLMv1., but also should not be involved here.

Oh, I lost the boat. I missed it with [openshift/operator-framework-olm](https://github.com/openshift/operator-framework-olm), which is OLMV0. Sorry, thank you for correcting it.

openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/metrics/catalogd_metrics_monitor_servicemonitor.yaml
camilamacedo86: Good catcher. That cannot work without we set the port.

tmshort:
@azych this likely needs to be rebased.

openshift/catalogd/manifests/04-role-openshift-catalogd-catalogd-metrics-monitor-role.yml
```diff
@@ -0,0 +1,20 @@
+---
```
tmshort, line 1: Is this separator necessary?

tmshort:
/approve
/lgtm
Looks as though @joelanford's requested changes might block merging.

camilamacedo86 (commented):
I'm not sure how or where we can add E2E tests that won’t conflict with the bumper and are valid only for downstream OCP.

@dtfranz @tmshort might have a better idea on how to approach this. I’d also like to assess whether implementing this downstream-only test is worthwhile. If maintaining such a test exclusively downstream proves too difficult, then:

- a) It might not be worth the effort

OR
- b) It could still be valuable as a small E2E test—**serving as an example or a foundation for expanding test coverage in the future.**

Otherwise, I'm fine with it. So, /lgtm.

c/c @joelanford

tmshort:
Yeah, I noticed no tests... the best place might be part of the origin tests:
https://github.com/openshift/origin

azych:
+1 I'll be following up with a test in https://github.com/openshift/origin once merged

PR #253, opened 2025-01-30
UPSTREAM: <carry>: use projected volume for CAs to avoid subPath limitations

joelanford:


openshift/catalogd/kustomize/overlays/openshift/olmv1-ns/patches/manager_deployment_certs.yaml
```diff
@@ -24,4 +18,4 @@
   value: "--tls-key=/var/certs/tls.key"
 - op: add
   path: /spec/template/spec/containers/0/args/-
-  value: "--ca-certs-dir=/var/trusted-cas"
```
tmshort, line 27: Why are we changing the name here?

joelanford: Avoid confusion since trusted-ca-bundle and service-ca are combined?

joelanford: But I don't have a strong opinion here, especially if we're going to end up splitting these up and using separate flags for various CA usages.

tmshort: But they were combined before. The latest solution I have avoids the use of the flags...

PR #254, opened 2025-01-30
UPSTREAM: <drop> Ensure catalog caches are valid after restart

oceanc80:
This addresses an intermittent failure we've seen where operator-controller appears to not resolve catalogs that exist. We identified the issue coming from the catalog caches not being valid after a restart.

This PR adds the following to address this issue:
- On startup of the binary, always clear the catalog cache directory

The PR also includes some small refactoring and fixes:
- moves certain unpack functions to an util package and are now shared by catalog and bundle image unpacker
- if unpack path happens to be an existing file, we delete it and unpack the image

### DIFF
--- a/catalogd/cmd/catalogd/main.go
+++ b/catalogd/cmd/catalogd/main.go
@@ -63,6 +63,7 @@ import (
 	"github.com/operator-framework/operator-controller/catalogd/internal/storage"
 	"github.com/operator-framework/operator-controller/catalogd/internal/version"
 	"github.com/operator-framework/operator-controller/catalogd/internal/webhook"
+	"github.com/operator-framework/operator-controller/internal/util"
 )
 
 var (
@@ -257,8 +258,8 @@ func main() {
 		systemNamespace = podNamespace()
 	}
 
-	if err := os.MkdirAll(cacheDir, 0700); err != nil {
-		setupLog.Error(err, "unable to create cache directory")
+	if err := util.EnsureEmptyDirectory(cacheDir, 0700); err != nil {
+		setupLog.Error(err, "unable to ensure empty cache directory")
 		os.Exit(1)
 	}
 
--- a/internal/rukpak/source/util.go
+++ b/internal/rukpak/source/util.go
@@ -1,0 +1,86 @@
+package source
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+// SetReadOnlyRecursive sets directory with path given by `root` as read-only
+func SetReadOnlyRecursive(root string) error {
+	return filepath.WalkDir(root, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		fi, err := d.Info()
+		if err != nil {
+			return err
+		}
+
+		if err := func() error {
+			switch typ := fi.Mode().Type(); typ {
+			case os.ModeSymlink:
+				// do not follow symlinks
+				// 1. if they resolve to other locations in the root, we'll find them anyway
+				// 2. if they resolve to other locations outside the root, we don't want to change their permissions
+				return nil
+			case os.ModeDir:
+				return os.Chmod(path, 0500)
+			case 0: // regular file
+				return os.Chmod(path, 0400)
+			default:
+				return fmt.Errorf("refusing to change ownership of file %q with type %v", path, typ.String())
+			}
+		}(); err != nil {
+			return err
+		}
+		return nil
+	})
+}
+
+// UnsetReadOnlyRecursive unsets directory with path given by `root` as read-only
+func UnsetReadOnlyRecursive(root string) error {
+	return filepath.WalkDir(root, func(path string, d os.DirEntry, err error) error {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		if !d.IsDir() {
+			return nil
+		}
+		if err := os.Chmod(path, 0700); err != nil {
+			return err
+		}
+		return nil
+	})
+}
+
+// DeleteReadOnlyRecursive deletes read-only directory with path given by `root`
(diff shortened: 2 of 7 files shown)


joelanford:
We'll need a bug to hang this off of for the backport. It might make sense to combine with this a few other improvements @tmshort and I are making that also need backports.

oceanc80:
> We'll need a bug to hang this off of for the backport. It might make sense to combine with this a few other improvements @tmshort and I are making that also need backports.

Do we want a new bug or to use one of the existing ones?

catalogd/cmd/catalogd/main.go
```diff
@@ -63,6 +63,7 @@ import (
 	"github.com/operator-framework/operator-controller/catalogd/internal/storage"
 	"github.com/operator-framework/operator-controller/catalogd/internal/version"
 	"github.com/operator-framework/operator-controller/catalogd/internal/webhook"
+	"github.com/operator-framework/operator-controller/internal/util"
```
azych, line 66: would it be possible to [avoid introducing `util` and have the new functions located in a more meaningful place(s)](https://100go.co/#creating-utility-packages-13)?

internal/rukpak/source/util.go
```diff
@@ -0,0 +1,86 @@
...
+func DeleteReadOnlyRecursive(root string) error {
+	if err := UnsetReadOnlyRecursive(root); err != nil {
+		return fmt.Errorf("error making directory writable for deletion: %w", err)
+	}
+	return os.RemoveAll(root)
+}
+
+// IsImageUnpacked checks whether an image has been unpacked in `unpackPath`.
+// If true, time of unpack will also be returned. If false unpack time is gibberish (zero/epoch time).
+// If `unpackPath` is a file, it will be deleted and false will be returned without an error.
```
azych, line 73: can you share more context for this change? Currently, the code is `panic`ing if `unpackPath` is not a directory, so it seems like a significant logic change to go to silent removal

internal/rukpak/source/util.go
```diff
@@ -0,0 +1,86 @@
...
+func IsImageUnpacked(unpackPath string) (bool, time.Time, error) {
+	unpackStat, err := os.Stat(unpackPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return false, time.Time{}, nil
+		}
+		return false, time.Time{}, err
+	}
```
azych, line 81: nit: can be flatten with:
```go
if errors.Is(err, os.ErrNotExist) {}
if err != nil {}
```

internal/util/fs.go
```diff
@@ -0,0 +1,23 @@
...
+func EnsureEmptyDirectory(path string, perm fs.FileMode) error {
+	entries, err := os.ReadDir(path)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	for _, entry := range entries {
+		if err := os.RemoveAll(filepath.Join(path, entry.Name())); err != nil {
+			return err
+		}
+	}
```
azych, line 21: it should be possible to make it more concise and readable by replacing lines 13-21 with just:
```go
if !os.Stat(path).IsDir() {
   return fmt.Errorf("path %q is not a directory", path) // or similar 
}
err := os.RemovAll(path)
...

File: internal/rukpak/source/util.go [hunk 5]
```diff
@@ -0,0 +1,86 @@
...
+		}
+		if !d.IsDir() {
+			return nil
+		}
+		if err := os.Chmod(path, 0700); err != nil {
+			return err
+		}
+		return nil
+	})
+}
```
azych, line 61: IMO it should be possible to extract `filepath.WalkDir(root, func(path string, d os.DirEntry, err error) error` that both `UnsetReadOnlyRecursive` and `SetReadOnlyRecursive` use to a private helper function with signature being something like:

```go
setModeRecursive(root, fileModeToSet, dirModeToSet)
```

and then call it in those public functions to avoid duplication

internal/rukpak/source/util.go
```diff
@@ -0,0 +1,86 @@
...
+		if errors.Is(err, os.ErrNotExist) {
+			return false, time.Time{}, nil
+		}
+		return false, time.Time{}, err
+	}
+	if !unpackStat.IsDir() {
+		return false, time.Time{}, os.Remove(unpackPath)
+	}
+	return true, unpackStat.ModTime(), nil
+}
```
azych, line 86: would be nice to have unit tests for those public functions

internal/util/fs.go
```diff
@@ -0,0 +1,23 @@
...
+	return os.MkdirAll(path, perm)
+}
```
azych, line 23: would be nice to have unit test for this

catalogd/cmd/catalogd/main.go
perdasilva: Do you have some suggestions? As always, naming is a PITA... I'm struggling to find a good name for a package that contains common shared utility functions used by various other packages in the project XD should we just call the package `fs`?

azych: if the functions / utils deal with filesystem then `fs` makes sense as package name to me

internal/util/fs.go
perdasilva: The problem is, path can point to a pod mount, in which case os.RemoveAll will fail. That's why we step through and delete each of the entries, rather than nuke the whole directory =S

internal/rukpak/source/util.go
perdasilva: agreed - this was a bit rushed because of the component readiness issues. I'm addressing all these issues in an upstream PR. Thank you for holding this line!

internal/rukpak/source/util.go
perdasilva: nice! let me do that

internal/util/fs.go
azych: thanks for the explanation, maybe a short doc comment for people potentially wanting to refactor it in the future?

PR #261, opened 2025-02-04
debug

azych:


azych:
this served its debugging purpose to find a resolution and use it in https://github.com/openshift/operator-framework-operator-controller/pull/252

PR #265, opened 2025-02-10
NO-ISSUE: Invalidate the containers/images cache

tmshort:
This is partial downstream of upstream PR 1554, but includes both catalogd and operator-controller


joelanford:
/hold

I'm not sure OCPBUGS-48795 is the correct bug for this change.

PR #267, opened 2025-02-13
NO-ISSUE: Remove vet from openshift verify

tmshort:
The `vet` target was removed upstream.


grokspawn (approved):

grokspawn:
removal of a make target invocation shouldn't ever be able to impact anything else.

PR #283, opened 2025-03-03
OCPBUGS-52230: PSA Labels version which should pin the version instead of use latest

camilamacedo86:
OLMv1 should avoid setting Pod Security Admission (PSA) labels to latest. Instead, it should pin the PSA version to match the Kubernetes API version specified in the go.mod file.​

Upstream change: https://github.com/operator-framework/operator-controller/pull/1828

### DIFF
--- a/openshift/Makefile
+++ b/openshift/Makefile
@@ -17,6 +17,19 @@ verify: ## Run downstream-specific verify
 manifests: $(KUSTOMIZE) $(YQ)
 	$(DIR)/operator-controller/generate-manifests.sh
 	$(DIR)/catalogd/generate-manifests.sh
+	make update-k8s-values
+
+# Minor Kubernetes version to build against derived from the client-go dependency version
+KUBE_MINOR ?= $(shell cd $(DIR)/.. && GOFLAGS=-mod=mod go list -m k8s.io/client-go | cut -d" " -f2 | sed -E 's/^v0\.([0-9]+)\.[0-9]+.*$$/1.\1/')
+
+.PHONY: update-k8s-values # HELP Update PSA labels in config manifests with Kubernetes version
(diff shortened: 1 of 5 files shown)


jianzhangbjz:
Test pass, details: https://issues.redhat.com/browse/OCPBUGS-52230
/label qe-approved
/lgtm

tmshort:
So, `make update-k8s-psa` is something we may have to do periodically?

camilamacedo86:
> So, make update-k8s-psa is something we may have to do periodically?

When we update the supported Kubernetes version (which happens when we bump `controller-runtime`—as each OCP release typically aligns with a `controller-runtime` version that upgrades to Kubernetes 1.X or its OCP equivalent), **running `make manifest` will automatically update the related files.**

We also validate this through `make verify`. So, if for any reason we bump the project's Kubernetes version (which doesn’t happen often), the verification step will fail, helping us catch it early. As agreed in [this discussion](https://redhat-internal.slack.com/archives/C06KP34REFJ/p1739880491760029), we use the Kubernetes API version specified in `go.mod`. In **99% of cases**, this should be correct, as it's rare that the Kubernetes API version doesn't match the OCP release.

xingxingxia:
Per discussion with Joe, we are fine to leave our namespace manifest with latest since it doesn't appear hypershift is using it, "latest" means the-version-of-the-control-plane-that-is-running-psa-logic-for-this-namespace. So these pinning PRs appear not needed.

jianzhangbjz:
Based on https://redhat-internal.slack.com/archives/C06KP34REFJ/p1741224828143489?thread_ts=1739880491.760029&cid=C06KP34REFJ, we should use the latest instead of the pinned version.
/remove-label qe-approved

PR #284, opened 2025-03-03
OCPBUGS-52234: [release-4.18] UPDATE PSA labels for OCP manifests with k8s version

camilamacedo86:
This change is the same as: https://github.com/openshift/operator-framework-operator-controller/pull/283 but adapted for the old structure. See that I need to solve here the generate script as well to be able to test it out on mac os

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,20 @@ export IMAGE_TAG
 
 IMG := $(IMAGE_REPO):$(IMAGE_TAG)
 
+# Extract Kubernetes client-go version used to set the version to the PSA labels, for ENVTEST and KIND
+ifeq ($(origin K8S_VERSION), undefined)
+K8S_VERSION := $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed -E 's/^v0\.([0-9]+)\.[0-9]+$$/1.\1/')
+endif
+
+# Ensure ENVTEST_VERSION follows correct "X.Y.x" format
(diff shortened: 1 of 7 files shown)


tmshort:
So, `make update-k8s-psa` is something we may have to do periodically?

camilamacedo86:
> So, make update-k8s-psa is something we may have to do periodically?

It's very unlikely that we'll need to update the Kubernetes version again. However, we're keeping the same approach as before to keep things simple and ensure we don't forget to bump the Kubernetes version for each OCP release.

As agreed in [this discussion](https://redhat-internal.slack.com/archives/C06KP34REFJ/p1739880491760029), we use the Kubernetes API version specified in `go.mod`. In **99% of cases**, this should be correct, as it's rare that the Kubernetes API version doesn't match the OCP release.

🔹 When we run `make manifest`, it updates the necessary files.  
🔹 This is validated in `make verify`.  
🔹 The Kubernetes version is **only updated** when we bump `controller-runtime` to a new major version that supports a newer Kubernetes version.  

This keeps everything clean and easy to maintain! 🚀

jianzhangbjz:
Based on https://redhat-internal.slack.com/archives/C06KP34REFJ/p1741224828143489?thread_ts=1739880491.760029&cid=C06KP34REFJ, we should use the latest instead of the pinned version. So, we don't need this change any more.
/close

PR #288, opened 2025-03-10
OCPBUGS-52282: UPSTREAM: <carry>: Make /etc/docker optional

tmshort:
Mount /etc/docker indirectly via symlink

### DIFF
--- a/openshift/catalogd/generate-manifests.sh
+++ b/openshift/catalogd/generate-manifests.sh
@@ -19,6 +19,8 @@ declare -A IMAGE_MAPPINGS
 IMAGE_MAPPINGS[kube-rbac-proxy]='${KUBE_RBAC_PROXY_IMAGE}'
 # shellcheck disable=SC2016
 IMAGE_MAPPINGS[manager]='${CATALOGD_IMAGE}'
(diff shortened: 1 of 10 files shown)


jianzhangbjz:
It seems this PR doesn't work, see details: https://issues.redhat.com/browse/OCPBUGS-52282

jianzhangbjz:
Hi @tmshort , this fixed PR works for this bug. However, it introduced the `x509` error since no `/etc/docker/certs.d` softlink there, details in https://issues.redhat.com/browse/OCPBUGS-52282

joelanford:
To me, this feels like a lot more moving parts than necessary. Before we merge, we should have a conversation with the image registry and node teams to figure out what the best path forward is.

I wonder if we could have the OS image or MCO create an empty `/etc/docker` directory.

tmshort:
@joelanford presumably, you know the people to talk to? I'm ok with or without this; at least we discovered a dependency (which might be another solution)

tmshort:
Closing in favor of 4.18 changes.
Also, forgot to mount host `/etc` as `/hostetc`...

PR #301, opened 2025-03-24
OCPBUGS-52282: Make /etc/docker optional

tmshort:
Mount /etc/docker indirectly via symlink

This is a 4.18-specific fix to deal with /etc/docker not existing on the node. This can occur when the ImageRegistry capability is not enabled. MCO will be updated to create /etc/docker on the node in 4.17 and later. This will fix 4.17.old (no MCO fix) -> 4.18.new.


tmshort:
Likely not needed, so closing.

PR #314, opened 2025-04-14
OCPBUGS-54877: UPSTREAM: <drop>: Fix log initialization (#1917)

anik120:
Use proper flags and remove use of logrus and klog


tmshort:
No, we want this to be an `UPSTREAM: <drop>:` commit

PR #320, opened 2025-04-17
Catalog sync second poc

camilamacedo86:


openshift/catalog-sync/pkg/extract/extract.go
```diff
@@ -0,0 +1,184 @@
...
+
+	destRef, err := layout.ParseReference(fmt.Sprintf("%s:%s", tmpDir, "v1"))
+	if err != nil {
+		return fmt.Errorf("parse layout ref: %w", err)
+	}
+
+	policy := &signature.Policy{
+		Default: []signature.PolicyRequirement{
+			signature.NewPRInsecureAcceptAnything(),
+		},
```
joelanford, line 76: We should not use the "insecure accept anything" policy. Instead, we should load the policy from the default filesystem location. That may mean mounting `/etc/containers` from the node, similar to what we do in catalogd and operator-controller.

openshift/catalog-sync/pkg/extract/extract.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return fmt.Errorf("create policy context: %w", err)
+	}
+	defer policyCtx.Destroy()
+
+	// Required to allow run the tests on Mac
+	sysCtx := &types.SystemContext{
+		OSChoice:           "linux",
+		ArchitectureChoice: "amd64",
+	}
```
joelanford, line 88: Pre-disposing `containers/image` to a particular OS and arch will interfere with our ability to evaluate the image at the OCI level.

We want to actually evaluate the root reference. We don't want `containers/image` automatically de-referencing a manifest list/index and giving us just one of the underlying manifests.

This reminds me: for each root reference, we:
1.  expect it to be a manifest list or index (I'm not sure which, we should figure that out and then require either manifest list or index, not either).
2. will have N filesystem and FBC checks to run, where N is the number of manifests in the manifest list/index.

openshift/catalog-sync/pkg/extract/extract.go
```diff
@@ -0,0 +1,184 @@
...
+
+	for _, layer := range mf.LayerInfos() {
+		rc, _, err := src.GetBlob(ctx, layer.BlobInfo, nil)
+		if err != nil {
+			return fmt.Errorf("get layer: %w", err)
+		}
+		if err := untarLayer(rc, fsDir); err != nil {
+			rc.Close()
+			return fmt.Errorf("extract layer: %w", err)
+		}
```
joelanford, line 135: We should use an existing libraries that are designed specifically for extracting image layers here.
- For automatic decompression: https://github.com/containers/image/blob/e2534e0da6eced91be073195189f2394ab72923a/pkg/compression/compression.go#L160
- For applying layers correctly: perhaps https://github.com/containers/storage/blob/29f052f570bb3c2839a55448faeb792dd03ac90d/pkg/archive/archive.go#L1046. Another option is https://github.com/containerd/containerd/blob/1e8955d8b7c5ee9ec6c2cdb1e6eedff0c7088f9d/pkg/archive/tar.go#L141

openshift/catalog-sync/test/suite_test.go
```diff
@@ -0,0 +1,90 @@
...
+	{
+		Name:     "redhat",
+		ImageRef: "registry.redhat.io/redhat/redhat-operator-index:v4.18",
+		Checks: check.Checks{
+			ImageChecks:      check.AllImageChecks(),
+			FilesystemChecks: check.AllFilesystemChecks(),
+			CatalogChecks:    check.AllCatalogChecks(),
+		},
+	},
+}
```
joelanford, line 90: This construction makes it too easy to assign different checks to different images. I think what we want is:
- A slice of images
- A single definition of the checks.
- A loop over the images where we run the same check definition for each image.

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: I will give a look on that. 👍

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: I think we’re still able to handle manifest lists—we can loop through the items and run checks as expected.

This change was mostly to make tests work on Mac. Without setting the OS, I was getting this error:

```
err: <*errors.errorString>{
    s: "no image found in manifest list for architecture \"arm64\", variant \"v8\", OS \"darwin\"", <== see looking for darwin :-( 
}
```

I cleaned it up a bit and added a comment. Hope that’s okay!

Thanks!

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: Changed here to simplify

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: Hi Joe 👋

This change is scoped to dev/test use only. We're intentionally using `InsecureAcceptAnything` here to avoid setup issues on macOS and Docker, where policy.json is often missing.

I'm not exactly sure where we mount the policy in catalogd or operator-controller, but in this case, mounting it feels like unnecessary complexity IMHO. This code is just for testing — we’re pulling our own images, not running anything in prod — so I’d lean toward keeping it simple.

KISS principle applies here, I think 😊
WDYT? Can we agree on this?
I added a comment to clarify either .. thx for looking on this one.

openshift/catalog-sync/pkg/extract/extract.go
```diff
@@ -0,0 +1,180 @@
...
+	defer policyCtx.Destroy()
+
+	// Force image resolution to Linux platform to avoid OS mismatch errors
+	// during testing on Mac. Example error:
+	// err: <*errors.errorString>{
+	//     s: "no image found in manifest list for architecture \"arm64\", variant \"v8\", OS \"darwin\"",
+	// }
+	sysCtx := &types.SystemContext{
+		OSChoice: "linux",
+	}
```
camilamacedo86, line 96: @joelanford ^ See the comment and update to clarify it better.

openshift/catalog-sync/pkg/check/catalog.go
```diff
@@ -0,0 +1,143 @@
...
+			const configsDir = "configs"
+
+			entries, err := fs.ReadDir(fbcFS, configsDir)
+			if err != nil {
+				return fmt.Errorf("read configs dir: %w", err)
+			}
+
+			var errs []error
+
+			for _, entry := range entries {
```
joelanford, line 32: This code makes the assumption that catalogs will have subdirectories, each of which contain FBC. We (OLM) have generally not been prescriptive of this hierarchy AND when we've see other pipeline code that assumes this hierarchy, we've pointed it out as potentially problematic.

Since this check is about metadata, and not the directory structure, I think we should `declcfg.Load` the entire `configsDir`, and perform the check from there.

openshift/catalog-sync/pkg/check/catalog.go
```diff
@@ -0,0 +1,143 @@
...
+						}
+						if prop.Type == "olm.bundle.object" {
+							hasObject = true
+						}
+					}
+
+					if (hasCSV && hasObject) || (!hasCSV && !hasObject) {
+						errs = append(errs, fmt.Errorf("bundle %q in package %q has both or neither of olm.bundle.object / olm.csv.metadata", bundle.Name, bundle.Package))
+					}
+				}
```
joelanford, line 64: This isn't quite right for what we care about.

This set of checks is for 4.20. In 4.20, we expect:
- every channel head to have `olm.csv.metadata`.
- no bundle to have `olm.bundle.object`.

With my above comment in mind, I would suggest:
1. Getting a `declcfg.DeclarativeConfig` for the entire catalog
2. Converting that to a model (at which point you'll have a map of package names)
3. For each package:
   a. For each channel, get the channel head and check that `olm.csv.metadata` is present
   b. For each bundle, check that `olm.bundle.object` is NOT present.

joelanford: .... I scroll, and now I see the next check is close to what I'm suggesting. We don't need both checks. I think we can just have the one check for "expected operator metadata" that checks what I describe above.

openshift/catalog-sync/pkg/check/catalog.go
```diff
@@ -0,0 +1,143 @@
...
+				headBundleNames := map[string]bool{}
+				for _, pkg := range models {
+					for _, ch := range pkg.Channels {
+						head, err := ch.Head()
+						if err != nil {
+							continue
+						}
+						headBundleNames[head.Name] = true
+					}
+				}
```
joelanford, line 119: No need to build a map of channel head names and loop through every single bundle. `head` is a pointer to a `model.Bundle` object, which should give you direct access to the bundle properties to check against.

openshift/catalog-sync/pkg/check/catalog.go
```diff
@@ -0,0 +1,143 @@
...
+					for _, prop := range bundle.Properties {
+						if prop.Type == "olm.csv.metadata" {
+							hasCSV = true
+							break
+						}
+					}
+					if !hasCSV {
+						errs = append(errs, fmt.Errorf("head bundle %q in package %q is missing olm.csv.metadata", bundle.Name, bundle.Package))
+					}
+				}
```
joelanford, line 137: With the channel head loop above is checking for `olm.csv.metadata`, this loop can be updated to check that every bundle _lacks_ `olm.bundle.object`.

openshift/catalog-sync/pkg/check/image.go
```diff
@@ -0,0 +1,88 @@
...
+				return fmt.Errorf("unrecognized manifest type: %s", imgManifest.MediaType)
+			}
+			return nil
+		},
+	}
+}
+
+func ImageHasLabels(expectedLabels map[string]string) ImageCheck {
+	return ImageCheck{
+		Name: "HasExpectedImageLabels",
```
joelanford, line 48: I noticed this in my code when I was looking at it the other day... Would it make sense to include a stringified version of `expectedLabels` in the `Name`? That way if this check is created multiple times with different label sets, the `Name` will ultimately still distinguish which sets of labels were not found.

openshift/catalog-sync/pkg/check/image_check.go
```diff
@@ -0,0 +1,79 @@
...
+
+type Checks struct {
+	ImageChecks      []ImageCheck
+	FilesystemChecks []FilesystemCheck
+	CatalogChecks    []CatalogCheck
+}
+
+type ImageCheckFunc func(ctx context.Context, root ocispec.Descriptor, target oras.ReadOnlyTarget) error
+type FilesystemCheckFunc func(ctx context.Context, imageFS fs.FS) error
+type CatalogCheckFunc func(ctx context.Context, fbcFS fs.FS) error
```
joelanford, line 23: This should probably change to:
```suggestion
type CatalogCheckFunc func(ctx context.Context, declcfg.DeclarativeConfig) error
```

That way, we can have our framework convert to that type once, and then all catalog-based checks don't have to repeat this (costly) step.

openshift/catalog-sync/pkg/check/image_check.go
```diff
@@ -0,0 +1,79 @@
...
+func Check(ctx context.Context, store oras.ReadOnlyTarget, ociDir string, checks Checks) error {
+	desc, err := store.Resolve(ctx, "v1")
+	if err != nil {
+		return fmt.Errorf("resolve descriptor: %w", err)
+	}
+
+	unpackPath := filepath.Join(ociDir, "fs")
+	if _, err := os.Stat(unpackPath); err != nil {
+		return fmt.Errorf("expected extracted filesystem at %q: %w", unpackPath, err)
+	}
```
joelanford, line 58: Move this inside a check for `len(checks.FilesystemChecks) > 0` (and put the check running loop inside that conditional bloack as well)?

openshift/catalog-sync/pkg/check/image_check.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, check := range checks.FilesystemChecks {
+		if err := check.Fn(ctx, os.DirFS(unpackPath)); err != nil {
+			checkErrors = append(checkErrors, Error{CheckName: fmt.Sprintf("FS:%s", check.Name), Err: err})
+		}
+	}
+	for _, check := range checks.CatalogChecks {
+		if err := check.Fn(ctx, os.DirFS(unpackPath)); err != nil {
+			checkErrors = append(checkErrors, Error{CheckName: fmt.Sprintf("FBC:%s", check.Name), Err: err})
+		}
+	}
```
joelanford, line 76: Follow-up from my suggestion above about using `declcfg.DeclarativeConfig` in the CatalogCheck signature, do a similar `if len(checks.CatalogChecks) > 0` conditional, where we only convert to declcfg and run the loop if we have at least one check that needs it.

openshift/catalog-sync/test/suite_test.go
```diff
@@ -0,0 +1,52 @@
...
+	RunSpecs(t, "Catalog Validation Suite")
+}
+
+// images contains the list of images to be validated.
+var images = []string{
+	"registry.redhat.io/redhat/community-operator-index:v4.18",
+	"registry.redhat.io/redhat/redhat-marketplace-index:v4.18",
+	"registry.redhat.io/redhat/certified-operator-index:v4.18",
+	"registry.redhat.io/redhat/redhat-operator-index:v4.18",
+}
```
joelanford, line 27: For now it is okay for these to be hardcoded, but let's change these to 4.19 tags. We are technically now in 4.20 development window, but I don't think 4.20 catalogs have been bootstrapped.

As soon as this merges, we should open a PR to bump these to 4.20 (which will fail), and then we go hound the pipelines to get the 4.20 bootstrapping done.

As a follow-up to this PR, we should avoid maintaining this list separately. We already have a set of default catalogs in our openshift manifests, so it would be nice to derive this list from our existing default `ClusterCatalog` objects.

openshift/catalog-sync/test/suite_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+var _ = Describe("Catalog Image Validation Suite", func() {
+	for _, url := range images {
+		name := getImageNameFrom(url)
```
joelanford, line 31: Is there some reason we need to do the name remapping? It could be useful if the entire ref shows up in the test name.

openshift/catalog-sync/pkg/extract/extract.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		// To avoid permission errors faced
+		mask := os.FileMode(0755)
+		opts := &archive.TarOptions{
+			NoLchown:          true,
+			IgnoreChownErrors: true,
+			InUserNS:          true,
+			ForceMask:         &mask,
+		}
```
joelanford, line 154: It would be helpful for each of these options to have comments explaining their purpose.

openshift/catalog-sync/pkg/extract/extract.go
```diff
@@ -0,0 +1,180 @@
...
+
+// isPermError detects known non-critical permission errors
+// required to Linux/Mac compatibility
+func isPermError(err error) bool {
+	msg := err.Error()
+	return strings.Contains(msg, "permission denied") &&
+		(strings.Contains(msg, ".bash_logout") ||
+			strings.Contains(msg, ".pwd.lock") ||
+			strings.Contains(msg, "/afs") ||
+			strings.Contains(msg, "/etc/X11"))
```
joelanford, line 179: This seems fragile. If a single file in a layer has a problem like this, it seems like the rest of that layer's contents will be skipped, which isn't great. I think I saw something in `containers/storage/pkg/archive` that wraps a tar reader that allows the wrapper to make arbitrary changes to the tar header and to the contents of the files, and also allowing the entire entry to be skipped.

Ideally, we handle all files, directories, symlinks, and hardlinks with the necessary chown/chmod/unset xattrs, and then error if we encounter any other type (e.g. socket, block device, pipe, character device, etc.)

openshift/catalog-sync/test/suite_test.go
camilamacedo86: OK. I will update it and give a look

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: I will add them 👍

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: I tried to use the options from archive but I will give look.

openshift/catalog-sync/Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: test-catalog-sync
```
dtfranz, line 1: What's the reasoning behind adding a whole new Makefile? Does it not make sense to just add this to `openshift/Makefile`? It would make it easier to run the commands through the `openshift/release` CI if they're all using the same Makefile.

camilamacedo86: These tests aren't directly related to OLMv1 components or its release.

The idea behind placing them under the catalog-sync directory was to keep everything related to those tests self-contained. That way, if we ever need to move or remove them in the future, it’ll be easier to manage without impacting other parts of the project. It also helps with maintainability and clarity.

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: I added them one by one.
But the latest one seems to solve all needs.
So, I could clean up it all.

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: I could clean it all seems that the issues that I faced before are no longer more faced anymore
Thx.

openshift/catalog-sync/pkg/check/image.go
camilamacedo86: done

openshift/catalog-sync/pkg/check/catalog.go
camilamacedo86: Great. Good catcher.
Done 👍

openshift/catalog-sync/pkg/check/catalog.go
camilamacedo86: Done

openshift/catalog-sync/pkg/check/catalog.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+)
+
+// AllCatalogChecks returns a list of all catalog checks to be performed.
+func AllCatalogChecks() []CatalogCheck {
+	return []CatalogCheck{
+		CatalogHasValidMetadata(),
+		CatalogHasValidMetadataForChannelHeads(),
```
joelanford, line 15: Sorry, I think we flew past each other in yesterday's review I left in this area.

The metadata check should be:
1. All channel heads have `olm.csv.metadata`
2. _Nothing_ has `olm.bundle.object`

We have no opinion (yet) about whether non-channel heads have `olm.csv.metadata`

camilamacedo86: Hi @joelanford 

OK. I am changing for that. However, that does not seems the check that we have in : https://github.com/konflux-ci/build-definitions/blob/main/task/fbc-validation/0.1/fbc-validation.yaml#L236-L253

(Code shared ^ when we did our first meeting)

openshift/catalog-sync/pkg/extract/extract.go
camilamacedo86: Done.

openshift/catalog-sync/Makefile
camilamacedo86: On top of that, we might need to add more targets in the future.

openshift/catalog-sync/pkg/check/catalog.go
camilamacedo86: Neverless it is done as requested now 👍 So, addressed.

openshift/default-catalog-consistency/pkg/check/catalog.go
```diff
@@ -0,0 +1,77 @@
...
+			var errs []error
+			packages, err := declcfg.ConvertToModel(cfg)
+			if err != nil {
+				return fmt.Errorf("unable to covert: %w", err)
+			}
+
+			for _, pkg := range packages {
+				for _, ch := range pkg.Channels {
+					channelHead, err := ch.Head()
+					if err != nil {
```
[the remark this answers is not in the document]
camilamacedo86: Done 👍

grokspawn:
I know this is nitty, but since the tide config here is `merge-method:rebase` ([ref](https://github.com/openshift/release/blob/98084153d4c4bb5cbf036d4d081ecd17eb3f9f7e/core-services/prow/02_config/openshift/operator-framework-operator-controller/_prowconfig.yaml#L14C5-L14C61)), could you squash this down to one commit for portability?

perdasilva:
I'm a bit confused why we are adding these tests to our pipeline. This seems like something the catalog pipelines should validate and ensure. I worry we are adding extra responsibilities to this pipeline, which is not responsible for delivering catalogs. Do we want to block an operator-controller PR because a catalog is malformed?

openshift/default-catalog-consistency/test/validate/suite_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github/operator-framework-operator-controller/openshift/default-catalog-consistency/test/utils"
+)
+
+func TestSuite(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Validate Catalog Test Suite")
+}
+
+// images is a list of image references to be validated.
+var images = func() []string {
```
perdasilva, line 22: curious as to why you chose to panic instead of doing something like:

```
var _ = Describe("Check Catalog Consistency", func() {
    images, err := utils.LoadImagesFromCatalogs()
    Expect(err).ToNot(HaveOccurred())
```

or something like that

openshift/default-catalog-consistency/test/utils/utils.go
```diff
@@ -0,0 +1,73 @@
...
+	if _, err := os.Stat(catalogsPath); os.IsNotExist(err) {
+		return nil, fmt.Errorf("catalogs path %s does not exist", catalogsPath)
+	}
+
+	err := filepath.WalkDir(catalogsPath, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if d.IsDir() || (!strings.HasSuffix(d.Name(), ".yaml")) {
```
perdasilva, line 29: should we also include `.yml` to make sure we don't miss anything?

openshift/default-catalog-consistency/test/utils/utils.go
```diff
@@ -0,0 +1,73 @@
...
+			return nil
+		}
+
+		content, err := os.ReadFile(path)
+		if err != nil {
+			return fmt.Errorf("read %s: %w", path, err)
+		}
+
```
perdasilva, line 37: maybe we could ignore and log any content that doesn't contain, e.g. `Kind: ClusterCatalog` that way if any other yamls are commingled in the directory we don't error our unnecessarily?

openshift/default-catalog-consistency/test/utils/utils.go
```diff
@@ -0,0 +1,73 @@
...
+	"sigs.k8s.io/yaml"
+	"strings"
+
+	apiv1 "github.com/operator-framework/operator-controller/api/v1"
+)
+
+// LoadImagesFromCatalogs reads the catalogs from the files used and returns a list of image references.
+func LoadImagesFromCatalogs() ([]string, error) {
+	var images []string
+	catalogsPath := "../../../catalogd/kustomize/overlays/openshift/catalogs"
```
perdasilva, line 17: it could be worth making the the path a parameter to the function and configurable at the suite level maybe through env var or something like that. Just a thought/suggestion...

openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -0,0 +1,167 @@
...
+// UnpackImage pulls the image, extracts it to disk, and opens it as an OCI store.
+func UnpackImage(ctx context.Context, imageRef, name string) (res *Result, err error) {
+	tmpDir, err := os.MkdirTemp("", fmt.Sprintf("oci-%s-", name))
+	if err != nil {
+		return nil, fmt.Errorf("create temp dir: %w", err)
+	}
+
+	cleanup := func() { os.RemoveAll(tmpDir) }
+
+	if err := pullImage(ctx, imageRef, tmpDir); err != nil {
```
perdasilva, line 37: since we're cleaning up at every error, it could be worth to do something like:

```
store, err := func() {}()
if err != nil {
   os.RemoveAll(tmpDir)
   return err
}
```

just a thought - nothing that needs to be actioned

openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	// On Linux systems with tools like podman, buildah, or skopeo, a default policy.json may
+	// be available at ~/.config/containers or /etc/containers.
+	// However, this is not the case on mac/docker envs.
+	// To avoid unnecessary setup complexity, we explicitly skip policy validation here only for local dev.
+	// Load default policy from standard locations (~/.config/containers/policy.json or /etc/containers/policy.json)
+	// If LOCAL_ENV=true, use a local insecure policy file
+	if os.Getenv("LOCAL_ENV") == "true" {
+		policyPath := filepath.Join("..", "..", "hack", "policy.json")
```
perdasilva, line 88: is policy validation something that we'd also want to validate as part of these tests? e.g. to ensure they have been properly signed, etc....?

openshift/default-catalog-consistency/pkg/check/image.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+			defer reader.Close()
+
+			var img specsgov1.Image
+			if err := json.NewDecoder(reader).Decode(&img); err != nil {
+				return err
+			}
+
+			var errs []error
+			for expectedLabel, expectedValue := range expectedLabels {
```
perdasilva, line 93: do we care if the image has more labels than the ones we expect?

openshift/default-catalog-consistency/pkg/check/image.go
```diff
@@ -0,0 +1,111 @@
...
+			if err != nil {
+				return err
+			}
+			defer fetch.Close()
+			var m specsgov1.Manifest
+			if err := json.NewDecoder(fetch).Decode(&m); err != nil {
+				return err
+			}
+			switch m.MediaType {
+			case specsgov1.MediaTypeImageManifest, manifest.DockerV2Schema2MediaType:
```
perdasilva, line 76: we perform this validation in ImageIsValidManifest, do we want/have to do it again here?

perdasilva: and if we do, would it make sense to push the logic getting the manifest to some common function?

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -0,0 +1,40 @@
...
+	return []FilesystemCheck{
+		FilesystemHasDirectoriesWith(
+			"configs",
+			"tmp/cache",
+			"tmp/cache/pogreb.v1"),
+	}
+}
+
+// FilesystemHasDirectoriesWith checks if the specified paths exist and are directories in the image filesystem.
+func FilesystemHasDirectoriesWith(paths ...string) FilesystemCheck {
```
perdasilva, line 21: Maybe `FilesystemHasDirectories` or `FilesystemHasDirectoriesWithPaths`?

openshift/default-catalog-consistency/pkg/check/catalog.go
```diff
@@ -0,0 +1,122 @@
...
+
+// AllCatalogChecks returns a list of all catalog checks to be performed.
+func AllCatalogChecks() []CatalogCheck {
+	return []CatalogCheck{
+		CatalogHasNotBundleObject(),
+		CatalogHasValidMetadataForChannelHeads(),
+	}
+}
+
+// CatalogHasNotBundleObject verifies that any bundle has not olm.bundle.object.
```
perdasilva, line 19: I'm ussing the suggest changes button for some reason =S

```
// CatalogHasNoBundleObjectProperties verifies that no bundle has declares a olm.bundle.object property.
```

openshift/default-catalog-consistency/pkg/check/catalog.go
```diff
@@ -0,0 +1,122 @@
...
+							errs = append(errs,
+								fmt.Errorf("bundle %q in package %q is a head of channel and does not either "+
+									"olm.csv.metadata or olm.bundle.object set",
+									channelHead.Name,
+									channelHead.Package.Name))
+						}
+					}
+				}
+			}
+
```
perdasilva, line 118: I don't think it matters but in the previous method you have:

```
if len(errs) == 0 {
   return nil
}
```

openshift/default-catalog-consistency/pkg/check/image.go
```diff
@@ -0,0 +1,111 @@
...
+			default:
+				return fmt.Errorf("unrecognized manifest type: %s", imgManifest.MediaType)
+			}
+			return nil
+		},
+	}
+}
+
+// ImageHasExpectedLabelsWith checks if the image has the expected labels.
+func ImageHasExpectedLabelsWith(expectedLabels map[string]string) ImageCheck {
```
perdasilva, line 52: Maybe just `ImageHasLabels`?

perdasilva:
/lgtm

just left some suggestions and questions

openshift/default-catalog-consistency/test/validate/suite_test.go
camilamacedo86: Before, I used panic to make sure LoadImagesFromCatalogs() was actually called. I had issues because Expect doesn’t work outside Ginkgo blocks, and tests were ending without running anything.

But now I made it work in a better way, like you suggested—moving the load inside the Describe block. So I’m closing this one.

openshift/default-catalog-consistency/test/utils/utils.go
camilamacedo86: All the files we want have the `.yaml `suffix:

Screenshot 2025-05-07 at 22 33 42


But OK, I accept your suggestion — it's done ✅

openshift/default-catalog-consistency/test/utils/utils.go
camilamacedo86: Good call 👍

openshift/default-catalog-consistency/pkg/check/image.go
camilamacedo86: Changed for `ImageHasLabelsWith` 
Just keep the With to make clear that it has args.

openshift/default-catalog-consistency/pkg/check/catalog.go
camilamacedo86: Accepted 👍

openshift/default-catalog-consistency/pkg/check/filesystem.go
camilamacedo86: 👍 done

openshift/default-catalog-consistency/pkg/check/image.go
camilamacedo86: 👍 done

openshift/default-catalog-consistency/pkg/check/catalog.go
camilamacedo86: we do not need it.
removed from the previous one 👍

openshift/default-catalog-consistency/pkg/check/image.go
camilamacedo86: I don’t think we need to worry about the extra labels — what kind of issues could they actually cause?

As I see it, our focus should be on ensuring the mandatory fields are properly set.
That said, do you see any reason why we should handle this differently?

camilamacedo86: But looking your next comment I think we were missing

```
			if len(img.Config.Labels) == 0 {
				return errors.New("image has no labels at all")
			}
```

Added as well 👍

openshift/default-catalog-consistency/pkg/check/image.go
```diff
@@ -0,0 +1,111 @@
...
+			pairs = append(pairs, fmt.Sprintf("%s=%s", k, v))
+		}
+		sort.Strings(pairs)
+		name = fmt.Sprintf("%s[%s]", name, strings.Join(pairs, ","))
+	}
+
+	return ImageCheck{
+		Name: name,
+		Fn: func(ctx context.Context, root specsgov1.Descriptor, target oras.ReadOnlyTarget) error {
+			fetch, err := target.Fetch(ctx, root)
```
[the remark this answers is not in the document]
camilamacedo86: I improved this one give a look now let me know if that address your concerns.

openshift/default-catalog-consistency/pkg/check/image.go
camilamacedo86: Theoretically, we could call one check without the other, so I think it makes sense to keep the validation centralized and use it in both scenarios. That said, let me know if you think any further adjustments would be appropriate.

openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -0,0 +1,167 @@
...
+	store, err := oci.New(tmpDir)
+	if err != nil {
+		cleanup()
+		return nil, fmt.Errorf("open OCI store: %w", err)
+	}
+
+	return &Result{
+		Store:   store,
+		TmpDir:  tmpDir,
+		Cleanup: cleanup,
```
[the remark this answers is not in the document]
camilamacedo86: 👍 Done

openshift/default-catalog-consistency/pkg/extract/extract.go
camilamacedo86: 👍 refactored . Good catcher

openshift/default-catalog-consistency/pkg/extract/extract.go
camilamacedo86: Because I changed this one, and I think we need to discuss it, I will create a new comment with the code changed.

camilamacedo86:
Hi @tmshort 

The title is updated.
Could we move with it as it is, and work on follow-ups if needed?

perdasilva:
/lgtm

awesome work, and thank you!!!

openshift/default-catalog-consistency/hack/policy.json
joelanford, line 1: We don't need to actually commit this file. I would recommend doing this for now:
https://github.com/openshift/operator-framework-operator-controller/blob/a2fd7089db389282bb370b8ae483f5d2764fd454/internal/shared/util/image/pull.go#L263-L275

openshift/default-catalog-consistency/pkg/check/check.go
```diff
@@ -0,0 +1,127 @@
...
+
+	var (
+		// checkErrors store a list of errors that occurred during the checks.
+		checkErrors []error
+		// imageFsView store a filesystem view (fs.FS) rooted at that unpacked image path.
+		imageFsView fs.FS
+	)
+
+	// Get the image descriptors used in the image checks.
+	desc, err := store.Resolve(ctx, "v1")
```
joelanford, line 77: Where is `v1` coming from here? Why is that hardcoded? Should this be something that comes from the catalog image reference?

openshift/default-catalog-consistency/pkg/check/check.go
```diff
@@ -0,0 +1,127 @@
...
+	const (
+		fsDir      = "fs"
+		configsDir = "configs"
+	)
+
+	var (
+		// checkErrors store a list of errors that occurred during the checks.
+		checkErrors []error
+		// imageFsView store a filesystem view (fs.FS) rooted at that unpacked image path.
+		imageFsView fs.FS
```
joelanford, line 73: Nit: `imageFS`

openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// pullImage pulls an image and writes it to OCI layout at the specified directory
+func pullImage(ctx context.Context, imageRef, tmpDir string) error {
+	srcRef, err := docker.ParseReference("//" + imageRef)
+	if err != nil {
+		return fmt.Errorf("parse image ref: %w", err)
+	}
+
+	destRef, err := layout.ParseReference(fmt.Sprintf("%s:%s", tmpDir, "v1"))
```
joelanford, line 77: Ah, I see. We are hardcoding to "v1" here. To me, this is pretty confusing. Do we need a tag at all for the layout reference?

Here's how we do this in the pull implementation shared between catalogd and operator-controller: https://github.com/openshift/operator-framework-operator-controller/blob/a2fd7089db389282bb370b8ae483f5d2764fd454/internal/shared/util/image/pull.go#L123-L126

openshift/default-catalog-consistency/hack/policy.json
camilamacedo86: Done as suggested 👍

openshift/default-catalog-consistency/pkg/check/check.go
camilamacedo86: The code was refactored to use the shared reference away to get the canonicalRef.

openshift/default-catalog-consistency/pkg/check/catalog.go
```diff
@@ -0,0 +1,118 @@
...
+// Catalogs containing `olm.csv.metadata` properties must be served by `opm` binaries
+// with version at least `1.28.0`.
+func CatalogHasValidMetadataForChannelHeads() CatalogCheck {
+	return CatalogCheck{
+		Name: "CatalogHasValidMetadataForChannelHeads",
+		Fn: func(ctx context.Context, cfg declcfg.DeclarativeConfig) error {
+			var errs []error
+			packages, err := declcfg.ConvertToModel(cfg)
+			if err != nil {
+				return fmt.Errorf("unable to covert: %w", err)
```
joelanford, line 69:
```suggestion
				return fmt.Errorf("unable to convert: %w", err)
```

openshift/default-catalog-consistency/pkg/check/catalog.go
```diff
@@ -0,0 +1,118 @@
...
+							}
+						}
+
+						if !found {
+							errs = append(errs,
+								fmt.Errorf("bundle %q in package %q is a head of channel and does not either "+
+									"olm.csv.metadata or olm.bundle.object set",
+									channelHead.Name,
+									channelHead.Package.Name))
+						}
```
joelanford, line 110: I think I would drop all of this. While yes, having `olm.bundle.object` on the channel head makes it easier to migrate to `olm.csv.metadata`, I don't think our tests should be too concerned with _how_ to fix the problem. Let's just highlight _what_ the problems are.

That leaves us more open in the future to solve the "hows" in ways that don't come with built-in assumptions like needing to migrate from something that had `olm.bundle.object`.

openshift/default-catalog-consistency/Makefile
```diff
@@ -0,0 +1,5 @@
+.PHONY: test-catalog
+test-catalog: ## Run the set of tests to validate the quality of catalogs
+	WHAT=catalog \
```
joelanford, line 3: Where is `WHAT` used?

openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+
+		// Force image resolution to Linux platform to avoid OS mismatch errors
+		// during testing on Mac. Example error:
+		// err: <*errors.errorString>{
+		//     s: "no image found in manifest list for architecture \"arm64\", variant \"v8\", OS \"darwin\"",
+		// }
+		sysCtx := &types.SystemContext{
+			OSChoice: "linux",
+		}
```
joelanford, line 64: For follow-up:

It looks like what's actually happening right now is that `containers/image` is resolving and pulling an image index based on the tag we provide, and then automatically choosing a manifest from that index.

So that means that by the time we see a descriptor for the image, we are seeing just one manifest, not the index and _all_ of the manifests.

If we could see all of the manifests, then we'd be able to add more checks:
1. Is the root thing actually an index?
2. Does the index have manifests for the 4 architectures we support (amd64, arm64, ppc64le, s390x)?
3. Are the contents of the FBC and cache layers identical for those 4 manifests (could detect just by comparing layer digests)

openshift/default-catalog-consistency/pkg/check/catalog.go
camilamacedo86: My motivation to add this check here is to ensure we do not fail in this scenario.
If so, we have a bug in the tool or in the pipeline, which would lead us to reach out to the author to get the solution fixed.

Currently, we have no scenario that fails within. So, I would like to keep ensuring we have no regressions. If w have we can have further data to know how to deal with.

openshift/default-catalog-consistency/Makefile
camilamacedo86: I believe that it will be used in the periodics/OCP CI ecosystem identification, so I can confirm its usage when I am doing the next step to integrate. If I see that is not required I can circle back and remove. 
I just want to keep to facilitate the integration.

openshift/default-catalog-consistency/pkg/check/catalog.go
camilamacedo86: done

openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -0,0 +1,223 @@
...
+		canonicalRef, err := resolveCanonicalRef(ctx, srcRef, sysCtx)
+		if err != nil {
+			return nil, fmt.Errorf("resolve canonical ref: %w", err)
+		}
+
+		digestTag = canonicalRef.String()
+		destRef, err := layout.ParseReference(fmt.Sprintf("%s:%s", tmpDir, digestTag))
+		if err != nil {
+			return nil, fmt.Errorf("parse layout ref: %w", err)
+		}
```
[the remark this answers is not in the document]
camilamacedo86: Done now. 👍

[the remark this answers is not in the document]
camilamacedo86: Done now 👍

PR #319, opened 2025-04-16
OCPBUGS-52617: Bump jinja2 from 3.1.5 to 3.1.6 [release-4.17]

anik120:
Bumps [jinja2](https://github.com/pallets/jinja) from 3.1.5 to 3.1.6.
- [Release notes](https://github.com/pallets/jinja/releases)
- [Changelog](https://github.com/pallets/jinja/blob/main/CHANGES.rst)
- [Commits](https://github.com/pallets/jinja/compare/3.1.5...3.1.6)

---
updated-dependencies:
- dependency-name: jinja2 dependency-type: direct:production 
...

Signed-off-by: dependabot[bot] <support@github.com>
Co-authored-by: dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>

PR #328, opened 2025-04-25
OCPBUGS-52612: Bump jinja2 from 3.1.4 to 3.1.6 [Release-4.16]

MrSanketkumar:
Bump the jinja2 from 3.1.4 to 3.1.6

Backport #319

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ colorama==0.4.6
 cssselect==1.2.0
 ghp-import==2.1.0
 idna==3.7
-Jinja2==3.1.4
+Jinja2==3.1.6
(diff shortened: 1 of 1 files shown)


camilamacedo86:
We should not proceed with this task for the old release.
The dependency in question is used only in upstream documentation.
Therefore, there is no vulnerability in the downstream products related to this dependency that would require action.
Given this, the Jira should be closed as Not Affected.

I am moving forward and closing this one as not affected, but feel free to re-open if you see fit. 
See that I closed the JIRA with the above as Not a bug.

MrSanketkumar:
Thanks, for the confirmation. Hence closing the PR.

PR #345, opened 2025-05-16
NO-ISSUE: [DefaultCatalogTests]: Allow to pass auth path for docker credentials

camilamacedo86:
… "

PR #347, opened 2025-05-17
NO-ISSUE: <carry>: [DefaultCatalogTests]: Moving parse of ENVVAR to the caller (follow-up 345)

camilamacedo86:
Motivated by @joelanford  comments: https://github.com/openshift/operator-framework-operator-controller/pull/345#discussion_r2094101091

c/c @joelanford

### DIFF
--- a/openshift/default-catalog-consistency/test/validate/suite_test.go
+++ b/openshift/default-catalog-consistency/test/validate/suite_test.go
@@ -3,6 +3,8 @@ package validate
 import (
 	"context"
 	"fmt"
+	"github.com/containers/image/v5/types"
+	"os"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


camilamacedo86:
c/c @joelanford

openshift/default-catalog-consistency/test/validate/suite_test.go
```diff
@@ -31,7 +33,7 @@ var _ = Describe("Check Catalog Consistency", func() {
 			ctx := context.Background()
 			By(fmt.Sprintf("Validating image: %s", url))
 
-			extractedImage, err := extract.UnpackImage(ctx, url, name)
+			extractedImage, err := extract.UnpackImage(ctx, url, name, authPath)
```
joelanford, line 36: This PR as is solves my immediate concern, but I do have a nit suggestion to make it even better:

I'd suggest passing the entire `types.SystemContext` struct into `UnpackImage` rather than just `authFile`. That way, if we ever need to configure any other field on that struct, we don't have to change the `UnpackImage` signature again (and again).

joelanford (approved):

PR #355, opened 2025-05-21
NO-ISSUE: Change code implementation to allow run OCP CI and avoid permissions issues when we extract the layers.

camilamacedo86:


openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -158,23 +160,33 @@ func extractLayers(ctx context.Context, layoutPath, fsPath, tag string) error {
...
+				return true, nil
+			}))
+			if err != nil {
+				return fmt.Errorf("apply layer %d: %w", i, err)
+			}
+			return nil
+		}()
 		if err != nil {
-			return fmt.Errorf("apply layer %d: %w", i, err)
+			return fmt.Errorf("decompress layer %d: %w", i, err)
```
joelanford, line 186:
```suggestion
			return fmt.Errorf("apply layer %d: %w", i, err)
```

openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -158,23 +160,33 @@ func extractLayers(ctx context.Context, layoutPath, fsPath, tag string) error {
...
 		}
 	}
+
```
joelanford, line 189: Extraneous new line?

openshift/default-catalog-consistency/pkg/extract/extract.go
camilamacedo86: Good call. Most important was to ensure that it would work with the OCP CI.
It worked 🎉 
Doing a follow up for this small cleanups

PR #362, opened 2025-05-28
OCPBUGS-56818, OPRUN-3869: [Default Catalog Consistency Test] (feat) add check for bin/opm files in filesystem

camilamacedo86:
Checks if given paths exist and point to executable files or valid symlinks.
See that the execution of the new tests added can be checked via the CI job.

P.S.: Tested with previous releases as well. All worked fine. 

/cherrypick release-4.19 release-4.18 release-4.17

### DIFF
--- a/openshift/default-catalog-consistency/pkg/check/filesystem.go
+++ b/openshift/default-catalog-consistency/pkg/check/filesystem.go
@@ -5,11 +5,15 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"os"
+	"os/exec"
+	"path/filepath"
 )
 
 // AllFilesystemChecks returns a list of filesystem checks to be performed on the image filesystem.
 func AllFilesystemChecks() []FilesystemCheck {
 	return []FilesystemCheck{
+		FilesystemHasGoExecutables("bin/opm"),
 		FilesystemHasDirectories(
 			"configs",
 			"tmp/cache",
@@ -21,7 +25,7 @@ func AllFilesystemChecks() []FilesystemCheck {
 func FilesystemHasDirectories(paths ...string) FilesystemCheck {
 	return FilesystemCheck{
 		Name: "FilesystemHasDirectories" + fmt.Sprintf(":(%q) ", paths),
-		Fn: func(ctx context.Context, imageFS fs.FS) error {
+		Fn: func(ctx context.Context, imageFS fs.FS, _ string) error {
 			var errs []error
 			for _, path := range paths {
 				stat, err := fs.Stat(imageFS, path)
@@ -37,4 +41,81 @@ func FilesystemHasDirectories(paths ...string) FilesystemCheck {
 			return errors.Join(errs...)
 		},
 	}
+}
+
+// FilesystemHasGoExecutables checks that the given paths exist inside the extracted image filesystem.
+// It supports regular files and symlinks:
+//   - If it's a regular file, we just check that it exists.
+//   - If it's a symlink, we check that the link target exists.
+//
+// This is useful for verifying container images where binaries may be symlinks,
+// In our image builds (see:
+// https://github.com/openshift/operator-framework-olm/blob/082d59a819afc43b24e9ca23c531bdfc35418722/operator-registry.Dockerfile#L16-L19),
+// the actual binaries are in /bin/registry/*, and /bin/* just has symlinks that point to them.
+func FilesystemHasGoExecutables(paths ...string) FilesystemCheck {
+	return FilesystemCheck{
+		Name: "FilesystemHasGoExecutables" + fmt.Sprintf(":(%q)", paths),
+
+		// We use the `os` package instead of fs.FS because fs.FS does not support Lstat or Readlink,
+		// which are needed to detect and resolve symlinks correctly.
+		// Therefore, we are looking to real file paths under the unpacked image (in tmpDir/fs).
+		Fn: func(_ context.Context, _ fs.FS, tmpDir string) error {
+			var errs []error
+			root := filepath.Join(tmpDir, "fs")
+
+			for _, rel := range paths {
+				fullPath := filepath.Join(root, rel)
+
+				binaryPath, err := resolveImageSymlink(root, fullPath)
+				if err != nil {
+					errs = append(errs, fmt.Errorf("path %q: %w", rel, err))
+					continue
+				}
+
+				out, err := exec.Command("go", "version", "-m", binaryPath).CombinedOutput()
+				if err != nil {
+					errs = append(errs, fmt.Errorf("not a Go binary or unreadable metadata for %q: %v\n%s", rel, err, out))
+					continue
+				}
+			}
+
+			return errors.Join(errs...)
+		},
+	}
+}
+
+// resolveImageSymlink resolve the symlink target.
+// Example 1: if the target is absolute like "/bin/registry/opm",
+// that means it should exist under the image root at tmpDir/fs/bin/registry/opm.
+// Example 2: if the target is relative like "registry/opm" or "../registry/opm",
+// resolve it relative to the symlink’s own directory.
+func resolveImageSymlink(root, fullPath string) (string, error) {
+	fi, err := os.Lstat(fullPath)
+	if err != nil {
+		return "", fmt.Errorf("error stating file: %w", err)
+	}
+
+	if fi.Mode()&os.ModeSymlink == 0 {
+		return fullPath, nil
+	}
+
+	target, err := os.Readlink(fullPath)
+	if err != nil {
+		return "", fmt.Errorf("unreadable symlink: %w", err)
+	}
+
+	var resolved string
+	if filepath.IsAbs(target) {
+		// remove leading "/" and resolve from root
+		resolved = filepath.Join(root, target[1:])
+	} else {
+		// If it's not a symlink, we just check that the file exists.
+		resolved = filepath.Join(filepath.Dir(fullPath), target)
+	}
+
+	if _, err := os.Stat(resolved); err != nil {
+		return "", fmt.Errorf("symlink points to missing target %q: %w", resolved, err)
+	}
+
+	return resolved, nil
 }
--- a/openshift/default-catalog-consistency/pkg/extract/extract.go
+++ b/openshift/default-catalog-consistency/pkg/extract/extract.go
@@ -179,9 +179,9 @@ func extractLayers(ctx context.Context, layoutPath, fsPath, tag string) error {
 				hdr.Uid = os.Getuid()
 				hdr.Gid = os.Getgid()
 				if hdr.FileInfo().IsDir() {
-					hdr.Mode = 0700
+					hdr.Mode = 0755
 				} else {
-					hdr.Mode = 0600
+					hdr.Mode = 0755
 				}
 				return true, nil
 			}))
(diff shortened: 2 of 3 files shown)


camilamacedo86:
c/c @joelanford wdyt?

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -38,3 +41,78 @@ func FilesystemHasDirectories(paths ...string) FilesystemCheck {
...
+
+					if !filepath.IsAbs(target) {
+						target = filepath.Join(filepath.Dir(full), target)
+					}
+
+					// Get the info of the binary/target
+					tgtInfo, err := os.Stat(target)
+					if err != nil {
+						errs = append(errs,
+							fmt.Errorf("find symlink %q which points to the target %q that doesn't exist: %w",
```
bentito, line 97: Same as LStat, this is probably too specific about the reason, could be permissions, etc,

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -38,3 +41,78 @@ func FilesystemHasDirectories(paths ...string) FilesystemCheck {
...
+			// ExecPerm useful to check the executable permission
+			// 0o100 (owner), 0o010 (group), 0o001 (any)
+			const ExecPerm = 0o111
+
+			for _, rel := range paths {
+				full := filepath.Join(root, rel)
+
+				fi, err := os.Lstat(full)
+				if err != nil {
+					errs = append(errs, fmt.Errorf("not found %q: %w",
```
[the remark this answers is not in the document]
camilamacedo86: The test being used here is **FilesystemHasExecutables**.
The issue occurs when a file or symlink is missing at the specified path — in other words:

FilesystemHasExecutables .. PATH .. **not found**: **<error returned by the check>**

So far, the error message already includes everything we need: the path that failed, and the reason (e.g. missing, inaccessible, or not executable).

I'm not sure what additional information we should add here — it seems complete and clear as-is.

openshift/default-catalog-consistency/pkg/check/filesystem.go
camilamacedo86: 👍 Updated thank you.

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -38,3 +41,76 @@ func FilesystemHasDirectories(paths ...string) FilesystemCheck {
...
 	}
 }
+
+// FilesystemHasExecutables checks whether the specified paths refer to existing, accessible, and executable file.
+// It handles both regular files and symbolic links:
+//   - For regular files: checks that the file exists and is executable.
+//   - For symlinks: follows the link and checks that the target is an executable regular file.
+//
+// Any issues (missing file, broken symlink, non-executable file, etc.) will be reported as errors.
+func FilesystemHasExecutables(root string, paths ...string) FilesystemCheck {
```
joelanford, line 51: Maybe I'm missing something, but it looks like we are calling this function with `bin/opm` as `root` with no `paths`.

But the function doesn't really do anything unless there are paths to iterate.

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -38,3 +41,76 @@ func FilesystemHasDirectories(paths ...string) FilesystemCheck {
...
+	return FilesystemCheck{
+		Name: "FilesystemHasExecutables" + fmt.Sprintf(":(%q)", paths),
+
+		Fn: func(_ context.Context, _ fs.FS) error {
```
joelanford, line 55: Any checks of the filesystem would need a `fs.FS` to check against right? All of the file accesses that use functions from the `os` package need to be refactored against the fs.FS that we need to pass in here.

bentito: It seems like `imageFS fs.FS` parameter is currently unused here. The new function performs file operations directly on the host OS (using os.* calls based on its root and paths arguments) rather than operating on the provided imageFS abstraction. If this is always how it is intended to be used, I think it's no problem, but if it's meant to possibly work with, for instance, image filesystems, it might be a problem.

camilamacedo86: Thanks for the feedback! Just to clarify:

What we need is to make sure `bin/opm` works when called, otherwise oc mirror will face issues. It's always a symlink (to bin/registry/opm), **so we must support symlinks.**

**The Go fs.FS doesn't support checking or following symlinks — no Lstat, no Readlink. So, we can’t use it here.**

Since we're working with a real unpacked image on disk (`/tmp/oci-*`), using `os.* functions brings the correct result, I mean do the check properly. As far as I see, we're not breaking any abstraction — we’re checking the actual filesystem, which includes symlinks.

**Also, this is already tested in OCP/ci, so we know it works.** See that we are calling this test in the CI in this PR

I'll add a comment in the code to explain. If you have ideas to improve, I’m happy to hear them!

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -5,11 +5,14 @@ import (
...
 	"fmt"
 	"io/fs"
+	"os"
+	"path/filepath"
 )
 
 // AllFilesystemChecks returns a list of filesystem checks to be performed on the image filesystem.
 func AllFilesystemChecks() []FilesystemCheck {
 	return []FilesystemCheck{
+		FilesystemHasExecutables("bin/opm"),
```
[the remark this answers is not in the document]
camilamacedo86: Yep, I will do that 👍  Thank you.

openshift/default-catalog-consistency/pkg/check/filesystem.go
camilamacedo86: Sorry, with your ocmment I realized that I sent the wrong version of the solution fixed.
We are passing in the func, I change the interface.

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -38,3 +41,67 @@ func FilesystemHasDirectories(paths ...string) FilesystemCheck {
...
+// This check avoids failing when the file is present but not marked executable —
+// for example, after image unpacking in some CI environments.
+func FilesystemHasValidPathsOrLinks(paths ...string) FilesystemCheck {
+	return FilesystemCheck{
+		Name: "FilesystemHasValidPathsOrLinks" + fmt.Sprintf(":(%q)", paths),
+
+		// We use the `os` package instead of fs.FS because fs.FS does not support Lstat or Readlink,
+		// which are needed to detect and resolve symlinks correctly.
+		// Therefore, we are looking to real file paths under the unpacked image (in tmpDir/fs).
+		Fn: func(_ context.Context, _ fs.FS, tmpDir string) error {
```
camilamacedo86, line 64: @joelanford @bentito ^ see I hope that clarifies your previous comment.
The Go fs.FS doesn't support checking or following symlinks.

For all images we create a symlink for the opm.
I clarified it in the doc of the func as well. 

Since we're working with a real unpacked image on disk (`/tmp/oci-*`), using `os.* functions brings the correct result, I mean do the check properly. As far as I see, we're not breaking any abstraction — we’re checking the actual filesystem, which includes symlinks.

See when we pass a invalid path

```
 <check.Error>{
                  CheckName: "[FS]:FilesystemHasValidPathsOrLinks:([\"bin/invalid\" \"bin/registry/opm\"])",
                  Err: <*errors.joinError | 0x1400106b350>{
                      errs: [
                          <*fmt.wrapError | 0x14000452680>{
                              msg: "error stating \"bin/invalid\": lstat /var/folders/n4/j272tr6d7hq63mf7_skv6zr80000gn/T/oci-redhat-marketplace-index-353931148/fs/bin/invalid: no such file or directory",
                              err: <*fs.PathError | 0x140006f8cc0>{
                                  Op: "lstat",
                                  Path: "/var/folders/n4/j272tr6d7hq63mf7_skv6zr80000gn/T/oci-redhat-marketplace-index-353931148/fs/bin/invalid",
                                  Err: <syscall.Errno>0x2,
                              },
                          },
                      ],
                  },
              },
          ],
```

So, I think that is the best that we can do to verify this one

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -38,3 +41,67 @@ func FilesystemHasDirectories(paths ...string) FilesystemCheck {
...
+						// remove leading "/" and resolve from root
+						resolved = filepath.Join(root, target[1:])
+					} else {
+						// resolve relative path
+						resolved = filepath.Join(filepath.Dir(fullPath), target)
+					}
+
+					if _, err := os.Stat(resolved); err != nil {
+						errs = append(errs, fmt.Errorf("symlink %q points to missing target %q: %w", rel, resolved, err))
+					}
```
camilamacedo86, line 100: @joelanford @bentito 


Sadly, my fist try to check if is an executable did not work out:

```
$ ls -la /var/folders/n4/j272tr6d7hq63mf7_skv6zr80000gn/T/oci-community-operator-index-2618243349/fs/bin/registry/opm
-rw-------  1 camilam  staff  91558832 20 May 10:30 /var/folders/n4/j272tr6d7hq63mf7_skv6zr80000gn/T/oci-community-operator-index-2618243349/fs/bin/registry/opm
```

When we unpack the image we lose the permissions bits, see: https://github.com/openshift/operator-framework-operator-controller/blob/main/openshift/default-catalog-consistency/pkg/extract/extract.go#L181-L185

openshift/default-catalog-consistency/pkg/check/filesystem.go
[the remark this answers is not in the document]
bentito: 👍

openshift/default-catalog-consistency/pkg/check/filesystem.go
bentito: maybe it would be clearer to just not take the abstract FS as a parameter?

openshift/default-catalog-consistency/pkg/check/filesystem.go
```diff
@@ -38,3 +41,67 @@ func FilesystemHasDirectories(paths ...string) FilesystemCheck {
 		},
 	}
 }
+
+// FilesystemHasValidPathsOrLinks checks that the given paths exist inside the extracted image filesystem.
+// It supports regular files and symlinks:
+//   - If it's a regular file, we just check that it exists.
+//   - If it's a symlink, we check that the link target exists.
```
perdasilva, line 48: do we care if it points to a particular target or not? or only whether it exists?

perdasilva: at the same time, this feels a bit like testing implementation - should we rather just do something like execute opm --version and grep for some string (or something like that). I.e. it shouldn't really matter if its a symlink or not, just that it can be executed and that its opm...?

perdasilva: maybe doing something like: `go version -m bin/opm` ?

openshift/default-catalog-consistency/pkg/extract/extract.go
```diff
@@ -179,9 +179,9 @@ func extractLayers(ctx context.Context, layoutPath, fsPath, tag string) error {
 				hdr.Uid = os.Getuid()
 				hdr.Gid = os.Getgid()
 				if hdr.FileInfo().IsDir() {
-					hdr.Mode = 0700
+					hdr.Mode = 0755
```
perdasilva, line 182: maybe just:

```
hdr.Mode = 0755
```

now that we're assigning the same mode for files and dirs?

PR #366, opened 2025-06-02
OCPBUGS-56956, OPRUN-3914:  [Default Catalog Consistency Test]: fix junit output format to allow generate xml

camilamacedo86:
See: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_operator-framework-operator-controller/366/pull-ci-openshift-operator-framework-operator-controller-main-default-catalog-consistency/1929498213128081408/artifacts/default-catalog-consistency/default-catalog-tests/artifacts/junit_olm.xml

/cherrypick release-4.19 release-4.18 release-4.17

PR #374, opened 2025-06-04
 [Default Catalog Consistency Test]: Cherry-Picks 

camilamacedo86:
- Fix JUnit output format to allow generating XML: https://github.com/openshift/operator-framework-operator-controller/pull/366 (we need to change the output so that it can be integrated with Sippy in a way that allows them to raise alerts when things go wrong). Sippy understands this JUnit format. 
- Rename the tests suite for a name requested by TRT: https://github.com/openshift/operator-framework-operator-controller/pull/372
- Enable Catalog Tests since those are not passing in those tests and pipeline fix the issues: https://github.com/openshift/operator-framework-operator-controller/pull/371

### DIFF
--- a/openshift/default-catalog-consistency/.gitignore
+++ b/openshift/default-catalog-consistency/.gitignore
@@ -1,2 +1,5 @@
 # Binaries for programs and plugins
 bin/*
+
+# Output from tests
+*.xml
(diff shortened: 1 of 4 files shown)


camilamacedo86:
@oceanc80 We will need your help on this one :-)

tmshort:
Is the title correct? "4.17"?

PR #393, opened 2025-07-01
OPRUN-3890,OPRUN-3983: Upgrade CatalogSources to use images from OCP release 4.20

camilamacedo86:


camilamacedo86:
>It looks like (at least) the community-operator-index for 4.20 isn't available yet?

No, it is not :-( 
We are waiting for that.

grokspawn (approved):

tmshort:
`[sig-olmv1][OCPFeatureGate:NewOLM][Skipped:Disconnected] OLMv1 operator installation should block cluster upgrades if an incompatible operator is installed [Suite:openshift/conformance/parallel]` is failing, we should remove it from openshift/origin...

PR #395, opened 2025-07-02
OPRUN-3969: Add support for experimental manifests

tmshort:
Update the openshift kustomize configuration for both operator-controller and catalogd.

Update the manifest generation scripts to put the core generation code into a function (ignore-whitespace will help with the review), so that it can be called twice; once for standard, and once for experimental.

Move around some of the kustomization directives to
* Create a patch kustomization (Component) file and move the patch directives from olmv1-ns there. This allows it to be referenced from a different directory.
* Add a kustomization file for tusted-ca. This allows it to be referenced from a different directory.
* Move the setting of the namePrefix for operator-controller; this makes the generation compatible with upstream feature components.
* Define experimental kustomization files that reference existing components.
* Reference the correct CRDs (standard or experimental).
* Add references to upstream feature components into the experimental manifests.

This *will* add `--feature-gates` options from the upstream feature components to the experimental manifests. The cluster-olm-operator will strip those arguments from the deployments before adding the enabled feature gates.

Update the Dockerfiles to include the experimental manifests and a copy script (`cp-manifests`) into the image containers. The complexity of having multiple sets of manifests mean that the simple initContainer copy mechanism found in cluster-olm-operator is no longer sufficient.

This attempts to keep backwards compatibility with older versions of cluster-olm-operator, specifically by keeping the original (standard) manifests in the original location, and adding the experimental manifests in a new directory. The new `cp-manifests` script is used by newer versions of cluster-olm-operator.

The manifests are generated in a separate <drop> commit (which will get regenerated upon the bumper), this should help with the review.

### DIFF
--- a/openshift/catalogd.Dockerfile
+++ b/openshift/catalogd.Dockerfile
@@ -6,7 +6,9 @@ RUN make go-build-local
 FROM registry.ci.openshift.org/ocp/4.20:base-rhel9
 USER 1001
 COPY --from=builder /build/bin/catalogd /catalogd
+COPY openshift/catalogd/cp-manifests /cp-manifests
 COPY openshift/catalogd/manifests /openshift/manifests
+COPY openshift/catalogd/manifests-experimental /openshift/manifests-experimental
 
 LABEL io.k8s.display-name="OpenShift Operator Lifecycle Manager Catalog Controller" \
       io.k8s.description="This is a component of OpenShift Container Platform that provides operator catalog support."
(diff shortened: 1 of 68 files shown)


camilamacedo86 (commented):
/lgtm

It seems fine for me 👍 
Thank you a lot for the clarifications

camilamacedo86:
/lgtm cancel

@tmshort 

I think we need ensure that is rebase
and run make verify to properly update the files

See that is failing in the make verify now

tmshort:
OK, looks as though I need to rebase!

PR #403, opened 2025-07-11
UPSTREAM: <carry>: [OTE]: Add binary in the operator controller image to allow proper integration with OCP tests

camilamacedo86:


camilamacedo86:
@kuiwang02

kuiwang02:
/lgtm

@camilamacedo86 after it is merged and new the ocp payload is delivered (I mean you could see the green build which has this PR in https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/#4.20.0-0.nightly), the https://github.com/openshift/origin/pull/29949 will pass

kuiwang02:
for me, I am not sure if the command in https://github.com/openshift/origin/pull/29949#issuecomment-3060499475 is ok because the key point is that the ocp cluster must be installed with the ocp load which has the external binary.

but, after this PR is merged, the https://github.com/openshift/origin/pull/29949 will pass after the new ocp load is generated.

so, could merge this PR firstly.

openshift/operator-controller.Dockerfile
```diff
@@ -1,14 +1,21 @@
...
+       cp ./bin/olmv1-tests-ext /tmp/build/olmv1-tests-ext && \
+       gzip -f /tmp/build/olmv1-tests-ext
 
 FROM registry.ci.openshift.org/ocp/4.20:base-rhel9
 USER 1001
 COPY --from=builder /build/bin/operator-controller /operator-controller
-COPY openshift/operator-controller/cp-manifests /cp-manifests
+COPY --from=builder /tmp/build/olmv1-tests-ext.gz /usr/bin/olmv1-tests-ext.gz
 COPY openshift/operator-controller/manifests /openshift/manifests
-COPY openshift/operator-controller/manifests-experimental /openshift/manifests-experimental
```
kuiwang02, line 11: @camilamacedo86 I am not sure why you remove this line.  when we add binary into component image, no need to remove it. if there is other reason to remove it, please do it. or else, better to keep it.

camilamacedo86: My main branch had not it at the time.
Yes, we should keep

camilamacedo86: Fixed

kuiwang02:
@camilamacedo86 thanks for quick action.

/lgtm

PR #407, opened 2025-07-17
NO-ISSUE: Fix experimental manifest copying

tmshort:
The standard manifest was being copied rather than the experimental manifest. This meant that the expected feature-flags are not present. This is failing now that we are doing a check for those feature-flags.


tmshort:
/hold
This fixes the **techpreview** build. This must pass (or at least, must not fail due to OLMv1) for this to be considered a fix.
Please unhold when **techpreview** is happy.

dtfranz:
The techpreview job succeeded, so I'm removing the hold
/unhold
/lgtm

PR #409, opened 2025-07-17
OPRUN-3992: [OTE] Add tracking mechanism

camilamacedo86:


kuiwang02:
/lgtm


@camilamacedo86 the code of this PR is ok, but after this PR is merged, the PR https://github.com/openshift/release/pull/67210 will not work  because
1, verify-metadata need binary, while src image has no it.
2, actually you do not need verify-metadata because the update-metadata exit with non-zero. it means that if the update is not correct, the update-metadata will fail firstly. But you could keep verify-metadata if you would like.

so, if you want https://github.com/openshift/release/pull/67210 work, need to change 
```shell
make verify-metadata
```
to
```shell
make build
make verify-metadata
```
or
```shell
make build-update
```
or else, you will meet error "No such file or directory"

camilamacedo86:
Hi @kuiwang02 

Regards; https://github.com/openshift/operator-framework-operator-controller/pull/409#issuecomment-3094959024

No, we should not change as suggested. Let me explain:

- When we are developing tests, we do **not** want to update the metadata.  
We need to build the binary **without** touching metadata.  
This allows us to test things locally, safely. 

If we update metadata too early, we cannot rename things later — it becomes a mess.  
We should only update metadata **after** we are 100% sure about the new names and tests.

- When we are 100% fine with our tests then: At that point, we run `make build-update`.  
This command **will** update the metadata.  
It will also **fail** if there’s a duplicate name or other issues — which is good.

- How we ensure that the metadata is ALWAYS updated:
The CI will run `make metadata-verify`.  See: https://github.com/openshift/release/pull/67210
This ensures the metadata JSON is **always up to date**.  
And we don’t merge **any** PR where the code and metadata don’t match.

Hope that clarifies.

openshift/tests-extension/Makefile
```diff
@@ -67,3 +70,64 @@ fix-lint: $(GOLANGCI_LINT) #HELP Fix lint issues
...
 	GO_COMPLIANCE_POLICY="exempt_all" go build -ldflags "$(LDFLAGS)" -mod=vendor -o $(TOOLS_BIN_DIR)/olmv1-tests-ext ./cmd/...
+
+.PHONY: update-metadata
+update-metadata: #HELP Build and run 'update-metadata' to generate test metadata
+	$(TOOLS_BIN_DIR)/olmv1-tests-ext update --component openshift:payload:olmv1
+	$(MAKE) clean-metadata
+
+
+# A TestID is a unique and stable identity for a test.
+# It is made up of: Component (Product:Type:ComponentName:TestDescription)
```
perdasilva, line 81:
```suggestion
# It is made up of ':' separated tokens (Product:Type:ComponentName:TestDescription)
```

perdasilva: or maybe

```suggestion
# Follows the format <Product>:<Type>:<ComponentName>:<TestDescription>, e.g. ...
```

perdasilva:
My only suggestion would be to move all those comments into a README so its easier to find and follow. But that could be done in a follow-up.

openshift/tests-extension/Makefile
```diff
@@ -67,3 +70,64 @@ fix-lint: $(GOLANGCI_LINT) #HELP Fix lint issues
...
+# It is required to ensure unique TestID over time.
+# More info:
+# - https://github.com/openshift/enhancements/blob/master/enhancements/testing/openshift-tests-extension.md#test-id
+# - https://github.com/openshift-eng/ci-test-mapping
+#──────────────────────────────────────────────────────────────
+# How to rename a test?
```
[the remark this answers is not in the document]
camilamacedo86: I think we can have in both since it is a very odd caveat :-(

camilamacedo86:
/hold
I want to address your comments @perdasilva before get merge.
Thank you a lot. I will ping for another lgtm

openshift/tests-extension/Makefile
perdasilva: ah, I think I got confused with the of: and read it closer to " : component" and thought that's what was trying to be conveyed. It's still not exactly clear what is meant by

It is made up of: Component (is this a general use of the term "Component" or a specific use in the context of the OTE stuff?

kuiwang02:
> openshift/release#67210

@camilamacedo86 
In https://github.com/openshift/release/pull/67210, if you use `make verify-metadata`, the job tests-extension-verify will fail with error "No such file or directory" because of the reason I mentioned above. so, suggest you to change it to 
```shell
make build
make verify-metadata
```
or
```shell
make build-update
```

kuiwang02:
/lgtm

and need to update https://github.com/openshift/release/pull/67210 later

PR #415, opened 2025-07-25
OPRUN-3964: OTE add first test from openshift/origin olmv1.go

camilamacedo86:
This PR adds the setup to initialise the test environment and access the cluster config.  
It also includes the first test migrated from OpenShift/Origin (more will be added in follow-up PRs; I want to ensure that we have a good base before migrating all).

We added `extlogs` to improve test logs. It uses Ginkgo’s writer to show clear info and warnings,  
which helps when tests fail. **Since we will need to migrate the tests, it is an excellent opportunity to ensure that we have good logs to cover the scenarios.** 

The setup utilises the controller-runtime client, similar to our E2E tests.  
This lets us run the tests locally with KinD (with OLMv1 installed) or in CI on OpenShift.  
**Only a few parts are OCP-specific; we have only 3/4 of the features specific to OCP, while the rest is the same as used in upstream.** Therefore, most tests can run using Kind and locally, which facilitates the development and troubleshooting process. Furthermore, my vision for the future is to explore the possibility of reusing the tests created for our e2e tests within our organisation, rather than duplicating staff efforts

**This makes writing and running tests for OCP much easier and faster.**

c/c @tmshort @kuiwang02 @dtfranz @anik120 @grokspawn


openshift/tests-extension/README.md
```diff
@@ -15,6 +15,121 @@ They use the framework: https://github.com/openshift-eng/openshift-tests-extensi
...
+    "lifecycle": "blocking",
+    "duration": 26,
+    "startTime": "2025-07-26 05:42:26.553852 UTC",
+    "endTime": "2025-07-26 05:42:26.580263 UTC",
+    "result": "passed",
+    "output": ""
+  }
+]
+```
+
```
camilamacedo86, line 93: @kuiwang02 It’s okay to add it in a follow-up, but I’d like to document a comprehensive guide here on how to run the tests locally using openshift-bot and the payload.

Ideally something like:

Step 1

Step 2

Step 3
...

If you’re able to provide more details, I’d really appreciate it!
I’ll try it locally myself and then submit this guide in a follow-up PR.

Thanks!

openshift/tests-extension/test/olmv1.go
```diff
@@ -2,15 +2,99 @@ package test
...
+						crd.plural,
+						crd.group,
+						crdObj.Spec.Versions)
+				}
+				Expect(found).To(BeTrue(), fmt.Sprintf("Expected version %q in CRD %s.%s", v,
+					crd.plural, crd.group))
+			}
+		}
+	})
+})
```
camilamacedo86, line 78: Same test done in : https://github.com/openshift/origin/blob/48d56f2e5bb8bcf6c4560fb077c7b809e18a5c19/test/extended/olm/olmv1.go#L34-L74

(Migration - After we migrate for here and check that all is fine with Sippy we can remove from opc/origin)

openshift/tests-extension/test/env/cluster.go
```diff
@@ -0,0 +1,140 @@
...
+	discoveryClient := discovery.NewDiscoveryClientForConfigOrDie(cfg)
+	isOcp := detectOpenShift(discoveryClient)
+
+	scheme := runtime.NewScheme()
+	utilruntime.Must(apiextensionsv1.AddToScheme(scheme))
+	if isOcp {
+		utilruntime.Must(configv1.AddToScheme(scheme))
+	}
+
+	k8sClient, err := crclient.New(cfg, crclient.Options{Scheme: scheme})
```
camilamacedo86, line 76: We don’t need to use RunCommand("oc") as done in https://github.com/openshift/origin/blob/48d56f2e5bb8bcf6c4560fb077c7b809e18a5c19/test/extended/util/client.go#L165-L193 and ALL that complixities. 

Instead, we can use the controller-runtime client, just like we do in our e2e tests, to interact with the cluster.

This approach helps:

- Keep tests compatible with both oc and Kubernetes
- Simplify the codebase
- Avoid depending on CLI tooling

For reference, see how we do our e2e tests:
https://github.com/openshift/operator-framework-operator-controller/tree/main/test

camilamacedo86: Hi @kuiwang02 

It works like our current e2e tests and keeps things simple.
**If we find any test that really needs something different, we’ll handle it case by case.
Let’s keep it simple (KISS) and only add what’s truly needed for those special cases.**
That way, we avoid unnecessary complexity and keep our setup clean.


If we try to add everything we might need up front, we risk falling into the premature optimization trap.
That often leads to unnecessary complexity and code that's harder to maintain — without actually solving real problems.

openshift/tests-extension/test/env/cluster.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// configureQPS sets high QPS and burst values to avoid client-side throttling during tests.
+// This makes tests faster by allowing many API calls without delay.
+// It's mainly needed in serial tests, where slow or throttled requests can cause flakes.
+// The default limits (QPS=5, Burst=10) are too low for most test workloads.
+func configureQPS(cfg *rest.Config) *rest.Config {
+	cfg.QPS = 10000
+	cfg.Burst = 10000
+	cfg.RateLimiter = nil
```
camilamacedo86, line 138: See that ocp/origin client utils will create a TOM of complexities to create a fakeRateLimiter that return nill at the end : https://github.com/openshift/kubernetes/blob/b94367cabf3edea99cfd7ecf7e8bafdc908d83ca/staging/src/k8s.io/client-go/util/flowcontrol/throttle.go#L152-L154

openshift/tests-extension/test/env/cluster.go
```diff
@@ -0,0 +1,140 @@
...
+	return false
+}
+
+// configureQPS sets high QPS and burst values to avoid client-side throttling during tests.
+// This makes tests faster by allowing many API calls without delay.
+// It's mainly needed in serial tests, where slow or throttled requests can cause flakes.
+// The default limits (QPS=5, Burst=10) are too low for most test workloads.
+func configureQPS(cfg *rest.Config) *rest.Config {
+	cfg.QPS = 20
+	cfg.Burst = 50
```
camilamacedo86, line 137: See that we are using the same values used in the OCP/origin test utils client: 
	configCopy.QPS = 10000
	configCopy.Burst = 10000
	
	
	https://github.com/openshift/origin/blob/48d56f2e5bb8bcf6c4560fb077c7b809e18a5c19/test/extended/util/client.go#L1205-L1206

openshift/tests-extension/.openshift-tests-extension/openshift_payload_olmv1.json
```diff
@@ -8,5 +8,15 @@
...
+  },
+  {
+    "name": "[sig-olmv1][OCPFeatureGate:NewOLM] OLMv1 CRDs should be installed",
+    "labels": {},
+    "resources": {
+      "isolation": {}
+    },
+    "source": "openshift:payload:olmv1",
+    "lifecycle": "blocking",
+    "environmentSelector": {}
```
camilamacedo86, line 20: We need to ensure the same names:

https://sippy.dptools.openshift.org/sippy-ng/tests/4.20?filters=%257B%2522items%2522%253A%255B%257B%2522columnField%2522%253A%2522current_runs%2522%252C%2522operatorValue%2522%253A%2522%253E%253D%2522%252C%2522value%2522%253A%25227%2522%257D%252C%257B%2522columnField%2522%253A%2522variants%2522%252C%2522not%2522%253Atrue%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522never-stable%2522%257D%252C%257B%2522columnField%2522%253A%2522variants%2522%252C%2522not%2522%253Atrue%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522aggregated%2522%257D%252C%257B%2522columnField%2522%253A%2522current_flake_percentage%2522%252C%2522not%2522%253Atrue%252C%2522operatorValue%2522%253A%2522%253D%2522%252C%2522value%2522%253A%2522100%2522%257D%252C%257B%2522id%2522%253A99%252C%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522%255BOCPFeatureGate%253ANewOLM%255D%2520OLMv1%2520CRDs%2522%257D%255D%252C%2522linkOperator%2522%253A%2522and%2522%257D&sort=asc&sortField=net_improvement

Screenshot 2025-07-26 at 18 56 29

camilamacedo86:
/acknowledge-critical-fixes-only 

^ we are not in code freeze yet.
It is only to be more cautious.
It is only 1 test for OTE, which should not impact the release at all
Furthermore, we are testing it out without e2e:

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_operator-framework-operator-controller/415/pull-ci-openshift-operator-framework-operator-controller-main-e2e-aws-techpreview-olmv1-ext/1949167652710649856

To ensure that all is fine.
We need to merge this one to unblock the next steps.

openshift/tests-extension/cmd/main.go
```diff
@@ -137,22 +137,6 @@ func main() {
...
-	//
-	// openshift-tests handles this via:
-	// - SuiteWithKubeTestInitializationPreSuite()
-	//   -> calls DecodeProvider() and InitializeTestFramework()
-	//
-	// We'll need to add similar logic when we start add the tests here.
-	//
-	// References:
-	// - https://github.com/openshift/origin/blob/main/pkg/cmd/openshift-tests/run/flags.go#L53
-	// - https://github.com/openshift/origin/blob/main/pkg/clioptions/clusterdiscovery/provider.go#L100
```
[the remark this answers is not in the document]
kuiwang02: @camilamacedo86 
 the OTE  suggests this way https://github.com/openshift-eng/openshift-tests-extension/blob/main/cmd/example-tests/main.go#L64-L66. 

Because your code has 
```go
func Init() *TestEnv {
	if testEnv == nil {
		testEnv = initTestEnv()
	}
	return testEnv
}
```
to avoid to init cluster repeatedly

so, I do not find side effective with initing it in BeforeEach, not taking the way suggested by OTE.
So, we can go in this way to monitor if we meet issue late.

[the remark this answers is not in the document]
camilamacedo86: we might need raise that in the OTE channel
Because if we do as suggested, we may face issues with parallel tests. ( if not same instance and etc )

openshift/tests-extension/test/env/cluster.go
kuiwang02: @camilamacedo86 
Frankly, QE test cases are far more complex than the origin OLMv1 cases. They’re similar to the complex cases in other Origin modules that rely on Origin’s NewCLI, which in turn depends on k8s.io/kubernetes/test/e2e/framework and various other initializations. 
Because the Origin OLMv1 cases are relatively simple, the cluster initialization work you’ve done so far is sufficient—but it may only apply to those simple OLMv1 cases. When I eventually migrate the QE cases, I’ll revisit and adapt the initialization logic; that work will take quite some time. Right now I’m tied up with other priorities and don’t have the bandwidth to tackle it. 
Fortunately, your current work can proceed as planned.

Therefore:

/lgtm

And maybe I’ll adjust the init logic when migrating the QE cases in the future.
Thanks

PR #422, opened 2025-08-06
OPRUN-4058: (Add readme) OTE - How to test locally with OCP instances

camilamacedo86:


openshift/tests-extension/README.md
```diff
@@ -26,6 +26,32 @@ Use the environment variable `KUBECONFIG` to point to your cluster configuration
...
+export KUBECONFIG=~/.kube/cluster-bot.kubeconfig
+```
+
+3. Run the tests using the `olmv1-tests-ext` binary.
+
+**Example:**
+```shell
+./bin/olmv1-tests-ext run-suite olmv1/all
+```
+
```
anik120, line 54: Can this whole section be condensed to: 


>An Openshift Cluster with olmv1 enabled is required to run the tests. Once the `KUBECONFIG` env variable is pointing to the kubeconfig file saved locally (eg `~/.kube/cluster-bot.kubeconfig`), the tests can be run using the `olmv1-tests-ext` binary: 
```shell
./bin/olmv1-tests-ext run-suite olmv1/all
```

That way we're not making assumptions about how the reader is getting their OCP clusters (there are other ways than using cluster-bot), or getting into "how do you get a cluster" considerations at all in this section of the doc.

camilamacedo86: Hi @anik120,

Thanks for reviewing this. I think it's important to keep the detailed, step-by-step examples in this section. The goal is to make it as simple as possible for anyone to run the tests, regardless of their background, and the code examples help with that a lot.

I also believe we need to be clear about the `techpreview` requirement. The current examples help to clarify this.

Would you be open to keeping the more detailed version? 
This way, we can make sure the documentation is very clear and easy to follow.

Also, it might make more sense for you if you open the doc in the review tab and preview it.
You will see that we have generic info above, and we add examples with both. (OCP and k8s )

PR #426, opened 2025-08-08
NO-ISSUE: [OTE] Refac: refac helper and olmv1 test to create namespace instead to use pre-existent

camilamacedo86:
The refactoring will be required for the other tests to use the helper
Also, it ensures that we create a namespace specific to the test instead of using a pre-existing one to ensure that tests are isolated.

### DIFF
--- a/openshift/tests-extension/pkg/helpers/cluster_extension.go
+++ b/openshift/tests-extension/pkg/helpers/cluster_extension.go
@@ -2,25 +2,27 @@ package helpers
 
 import (
 	"context"
+	"fmt"
+	"time"
 
 	//nolint:staticcheck // ST1001: dot-imports for readability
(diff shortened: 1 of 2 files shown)


openshift/tests-extension/pkg/helpers/cluster_extension.go
```diff
@@ -2,25 +2,27 @@ package helpers
 
 import (
 	"context"
+	"fmt"
+	"time"
```
camilamacedo86, line 6: @kuiwang02 @tmshort 

It is an improvement:
- I just want to ensure that the tests are executed in a new namespace created for each test
- Also, I refac the helper since it will be used by other tests as well in a way that make easier 

Could you help me get this one merged?

PR #427, opened 2025-08-11
OPRUN-4068: OTE: rewrite the upgrade incompatible operator test

tmshort:
This test replaces the existing upgrade incompatible test. The main change is that operator and catalog bundles are created on-the-fly to support OCP 4.20. This means we are no longer dependent on public operators for this test.

This creates new bundles in the OCP ImageRegistry, this requires using a number of OCP APIs, including using a raw API URL to invoke the build. This is done by invoking an external k8s client (either `oc` or `kubectl`), and passing it a tarball of the bundle to be created. So, it can't be done by the golang k8sClient normally available (i.e. the create input is a tarball not a YAML file).

This introduces the use of go-bindata to store the bundle contents.

It also pulls in openshift image, buld and operator APIs.


openshift/tests-extension/.openshift-tests-extension/openshift_payload_olmv1.json
```diff
@@ -40,7 +40,7 @@
     "environmentSelector": {}
   },
   {
-    "name": "[sig-olmv1][OCPFeatureGate:NewOLM][Skipped:Disconnected] OLMv1 operator installation should block cluster upgrades if an incompatible operator is installed",
+    "name": "[sig-olmv1][OCPFeatureGate:NewOLM][Skipped:Disconnected] OLMv1 operator installation should fail to install a non-existing cluster extension",
```
camilamacedo86, line 43: We cannot change the names
If we do that we broke component readness

openshift/tests-extension/test/olmv1-incompatible.go
```diff
@@ -0,0 +1,500 @@
...
+		waitForOlmUpgradeStatus(ctx, operatorv1.ConditionFalse, ceName)
+
+		By("waiting for ClusterOperator Upgradeable to be false")
+		waitForClusterOperatorUpgradable(ctx, ceName)
+
+		By("finished")
+	})
+})
+
+func createClusterExtension(name, namespace, serviceaccount, bundle string) *olmv1.ClusterExtension {
```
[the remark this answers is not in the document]
tmshort: If you look, the code is waiting for the ClusterExtension to come up. I'm not using the current set of helper functions because they don't do _exactly_ what I want them to do. Basically, I want to know what the `unique` value is throughout the whole test (for _all_ my resources) and `unique` is not exposed.

See lines 182~189.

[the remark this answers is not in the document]
tmshort: The ClusterExtension is a non-namespace-scoped resource, and does not get cleaned up when when the namespace is deleted. So, there _is_ an explicit clean up of the non-namespace-scoped resources.
Please look at all the cleanup that does happen.

openshift/tests-extension/test/olmv1-incompatible.go
```diff
@@ -0,0 +1,500 @@
...
+
+		progressing := meta.FindStatusCondition(ce.Status.Conditions, olmv1.TypeProgressing)
+		g.Expect(progressing).ToNot(BeNil())
+		g.Expect(progressing.Status).To(Equal(metav1.ConditionTrue))
+
+		installed := meta.FindStatusCondition(ce.Status.Conditions, olmv1.TypeInstalled)
+		g.Expect(installed).ToNot(BeNil())
+		g.Expect(installed.Status).To(Equal(metav1.ConditionTrue))
+	}).WithTimeout(5 * time.Minute).WithPolling(1 * time.Second).Should(Succeed())
+}
```
camilamacedo86, line 408: Could we use instead:

https://github.com/openshift/operator-framework-operator-controller/blob/56f527af1025f3e1d5ff46dd58614b25cf6501ca/openshift/tests-extension/pkg/helpers/cluster_extension.go#L113-L131

So, we keep those centralised.

openshift/tests-extension/test/olmv1-incompatible.go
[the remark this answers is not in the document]
camilamacedo86: We can have only one Operator/Content per cluster.
If we install a cluster extension in one test, then try to install the same one in another test (not installed by OLM), it will fail :-( for now we use this one only here but I think we should probably re-use more of it :-) Since it is amazing 

Ideally, each test should delete everything it creates.  
We use unique names to avoid conflicts. (We need to overthing because of Component Readiness) And I know, some resources will not cause problems if left over, but they make troubleshooting harder when many are left behind.

**You did a fantastic job here and solved a big problem** 🚀  
In follow-up PRs, we’ll add more helpers. So, I propose

> If this passes CI 💥, we’ll merge it.  
> Later, we can move some parts to pkg as utils, which will also help QE provide tests.  
**All good to improve in future updates.**

openshift/tests-extension/.openshift-tests-extension/openshift_payload_olmv1.json
camilamacedo86: Yep, I validate thank you a lot !!!

openshift/tests-extension/test/olmv1-incompatible.go
tmshort: Yes, I modified the code to use this.

tmshort:
Updated to use `helper` routines, and tidied up the cleanup a bit, and renamed "bundle" to be "operator", which is a bit more precise.

openshift/tests-extension/pkg/helpers/cluster_extension.go
```diff
@@ -26,10 +26,12 @@ import (
 
 // CreateClusterExtension creates a ServiceAccount, ClusterRoleBinding, and ClusterExtension using typed APIs.
 // It returns the unique suffix and a cleanup function.
-func CreateClusterExtension(packageName, version, namespace string) (string, func()) {
+func CreateClusterExtension(packageName, version, namespace, unique string) (string, func()) {
 	ctx := context.TODO()
 	k8sClient := env.Get().K8sClient
-	unique := rand.String(4)
+	if unique == "" {
```
camilamacedo86, line 32: 👍

tmshort:
I added support to get the OCP version, so, now, we should never need to update this test!

PR #494, opened 2025-10-01
OPRUN-4080: Remove kustomize manifests from images and repo

tmshort:
Now that helm manifests are being used to dynamically generate the manifests, the pre-generated manifests are no longer needed. So, we can remove them from the repo and the images.

However, because we still want to verify the manifests are "good", we are still creating a "single-file" version of the manifests for verification purposes, and to allow us to see what changes are happening to the manifests (from upstream and/or downstream sources).


tmshort:
The techpreview's prior run had passed, but ran out of time (4h interrupt).

PR #500, opened 2025-10-07
NO-ISSUE: Update webhook ote tests to use latest webhook-operator

perdasilva:


openshift/tests-extension/test/olmv1-webhooks.go
```diff
@@ -0,0 +1,383 @@
+package test
```
camilamacedo86, line 1: All tests here are for olmv1
We never will add tests that are not for olmv1
So, could we keep the orginal name of the file so that we can easily review the changes?
What is new or not?

perdasilva: No, I just renamed the file to conform with the others.

perdasilva: I don't mind either way. It would just be good to have consistent naming of files in the package. I'll do a git mv.

openshift/tests-extension/test/webhooks.go
```diff
@@ -0,0 +1,383 @@
+package test
```
perdasilva: I've reverted for now. Maybe we can do a separate PR to rename everything one way or the other.

tmshort:
This should really have a ticket associated with it (and not NO-ISSUE)

openshift/tests-extension/test/webhooks.go
```diff
@@ -1,383 +1,383 @@
...
+    By("creating a ClusterRoleBinding to cluster-admin for the webhook operator")
+    operatorClusterRoleBindingName := fmt.Sprintf("%s-operator-crb", webhookOperatorInstallNamespace)
+    operatorClusterRoleBinding := helpers.NewClusterRoleBinding(operatorClusterRoleBindingName, "cluster-admin", saName, webhookOperatorInstallNamespace)
+    err = k8sClient.Create(ctx, operatorClusterRoleBinding)
+    Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("failed to create ClusterRoleBinding %s",
+        operatorClusterRoleBindingName))
+    helpers.ExpectClusterRoleBindingExists(ctx, operatorClusterRoleBindingName)
+
+    ceName := webhookOperatorInstallNamespace
+    ce := helpers.NewClusterExtensionObject("webhook-operator", "0.0.5", ceName, saName, webhookOperatorInstallNamespace)
```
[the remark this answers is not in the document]
camilamacedo86: @perdasilva any reason to hold?
If the tests does not pass, then we know that has an error and the CI will be blocked.

tmshort:
The `e2e-aws-techpreview-olmv1-ext` tests failed, and I suspect the openshift-e2e-aws-* tests will fail as well

PR #542, opened 2025-11-04
OPRUN-4212: [OTE]: Add webhook cleanup validation on extension uninstall

camilamacedo86:


tmshort:
/hold
Hold until  #546 merges, or this will cause another manual merge to occur!

PR #675, opened 2026-03-24
OCPBUGS-77358: Update golang.org/x/crypto to v0.48.0 [release-4.21]

tmshort:
Update to the latest version of golang.org/x/crypto and include any dependencies.


tmshort:
/close
The bug was reported against the rukpak image, which does not exist. The issue may not exist in operator-controller/catalogd, so I'm closing this.

PR #703, opened 2026-04-20
OPRUN-4577: OTE: Simplify by remove option to configure tests to run outside of OCP

camilamacedo86:
Initially, we considered designing OTE tests in a way that would allow them to run outside of OpenShift. However, almost all implemented tests are configured with IsOpenShift (just preflight and webhooks that are not), and the solution has only been validated against OCP. Therefore, this flexibility is no longer necessary. It seems that the costs of having it are not bringing any value. To keep things simple (KISS principle) and reduce unnecessary complexity, we should remove this option.

## Summary by CodeRabbit

* **Tests**
  * Tests now run only against OpenShift environments; OpenShift-specific capability and registry checks are enforced for all relevant test suites and specs.
  * Removed prior conditional skips—tests consistently perform OpenShift gating and proceed or skip based on detected OpenShift capabilities.

* **Documentation**
  * Local-run docs updated to require OpenShift; guidance for running on vanilla Kubernetes/Kind and related sections removed.

### DIFF
--- a/openshift/tests-extension/README.md
+++ b/openshift/tests-extension/README.md
@@ -2,7 +2,7 @@
 ========================
 
 This repository contains the OLMv1 tests for OpenShift.
-These tests run against OpenShift clusters and are meant to be used in the OpenShift CI/CD pipeline.
+These tests are designed to run exclusively on OpenShift clusters and are used in the OpenShift CI/CD pipeline.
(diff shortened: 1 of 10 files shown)


camilamacedo86:
c/c @bandrade

PR #712, opened 2026-05-04
OCPBUGS-84943: fix(test): drop blocking namespace-deletion wait between both-watch-modes scenarios

tmshort:
The both-watch-modes test loops over two scenarios (singlens, ownns) inside
a single It block and was blocking on full namespace deletion between them.
This caused flaky 300s timeouts on GCP techpreview clusters where master
nodes run at 94-99% CPU, which starves the namespace controller and makes
namespace termination arbitrarily slow.

The wait was not guarding anything real:
- EnsureCleanupClusterExtension already ensures the CE and CRD are gone;
  since CE deletion uses ForegroundPropagation, the ClusterObjectSet teardown
  must complete before the CE disappears, meaning all managed resources
  (Deployments, Services, etc.) are already deleted at that point.
- The singleown bundle installs no ValidatingWebhookConfiguration or
  MutatingWebhookConfiguration, so there is no webhook admission risk.
- Each scenario generates unique namespace names and CRD group suffixes via
  rand.String(4), so a terminating namespace from scenario 1 cannot collide
  with or interfere with scenario 2's resources.

Trigger both namespace deletions and proceed without waiting. The DeferCleanup
registrations that already exist will handle any residual cleanup after the
spec exits.

Fixes: OCPBUGS-84943

## Summary by CodeRabbit

* **Tests**
  * Speed up test iterations by removing per-iteration blocking waits for namespace deletion.
  * Rely on deferred cleanup handlers instead of explicit deletes between iterations.
  * Ensure teardown reliably targets the intended cluster-scoped resources by fixing cleanup invocation to use correct identifiers.

### DIFF
--- a/openshift/tests-extension/test/olmv1-singleownnamespace.go
+++ b/openshift/tests-extension/test/olmv1-singleownnamespace.go
@@ -13,7 +13,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/rand"
@@ -417,7 +416,7 @@ var _ = Describe("[sig-olmv1][OCPFeatureGate:NewOLMOwnSingleNamespace] OLMv1 ope
 				DeferCleanup(func() {
 					By(fmt.Sprintf("cleanup: deleting ClusterExtension %s", ce.Name))
 					_ = k8sClient.Delete(context.Background(), ce, client.PropagationPolicy(metav1.DeletePropagationForeground))
-					helpers.EnsureCleanupClusterExtension(context.Background(), ceName, nsName)
+					helpers.EnsureCleanupClusterExtension(context.Background(), packageName, crdName)
 				})
 
 				By(fmt.Sprintf("waiting for the ClusterExtension %s to be installed for %s scenario", ceName, sc.label))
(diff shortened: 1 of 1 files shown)


tmshort:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-e2e-gcp-ovn-techpreview 10

grokspawn (approved):

grokspawn:
I do question the rationale of considering a TPNU-only scenario a release blocker, but that's not this PR's fault.

openshift/tests-extension/test/olmv1-singleownnamespace.go
```diff
@@ -455,23 +454,15 @@ var _ = Describe("[sig-olmv1][OCPFeatureGate:NewOLMOwnSingleNamespace] OLMv1 ope
 				Expect(k8sClient.Delete(ctx, crb, client.PropagationPolicy(deletePolicy))).To(Succeed(), "failed to delete ClusterRoleBinding %q", crbName)
 				Expect(k8sClient.Delete(ctx, sa, client.PropagationPolicy(deletePolicy))).To(Succeed(), "failed to delete ServiceAccount %q", saName)
 
+				// Trigger namespace deletion and proceed without blocking. By this point
```
grokspawn: We actually could just rely on DeferCleanup at the end of the scenario, since there isn't resource collision. 
And for line 419, isn't the last parameter intended to be the CRD name, not the namespace?

Also, the comment at 447 is no longer correct (which I guess was before the resource parallelization, which might be the source of some CPU complexity but which obviates the need for serial namespace cleanup).

rashmigottipati (approved):
/lgtm

openshift/tests-extension/test/olmv1-singleownnamespace.go
tmshort: Claude said that "DeferCleanup" was less urgent than the rest of your comment.

grokspawn: I... see.  Blink twice if claude won't let you type any more.

tmshort: Updated

tmshort:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-e2e-gcp-ovn-techpreview 10

openshift/tests-extension/test/olmv1-singleownnamespace.go
tmshort: But... this seems to be causing failures now, so I don't think I can rely on just DeferCleanup.

tmshort:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-e2e-gcp-ovn-techpreview 10

grokspawn (approved):

tmshort:
Looks as though e2e-gcp-ovn-upgrade _will_ pass.

tmshort:
/hold
Until we look at the two `/payload-aggregate` jobs.

tmshort:
/unhold
The most recent aggregate all passed.

rashmigottipati (approved):
/lgtm

PR #745, opened 2026-06-03
OPRUN-4392,OPRUN-4393: Add OLMv1 progress deadline QE tests

bandrade:
## Summary

- Automates OCP-88331 and OCP-88332 in the OpenShift OLMv1 QE suite.
- Adds a focused ClusterExtension progress-deadline spec that builds in-cluster bundle and catalog images for successful and failing rollout versions.
- Verifies ProgressDeadlineExceeded on an initial failed rollout and ProbeFailure while upgrading to a bad revision under the NewOLMBoxCutterRuntime feature gate.

## Validation

- `go test -mod=vendor ./test/qe/specs`
- `make build`
- `./bin/olmv1-tests-ext list -o names | rg '88331|88332'`\n- `make verify GO=/Users/bandrade/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.7.darwin-arm64/bin/go`

## Summary by CodeRabbit

* **Tests**
  * Added a new Ginkgo test covering ClusterExtension rollout behavior when persistent bundle/controller image errors occur, including progress-deadline timeouts and the expected status/reason propagation.
  * Added coverage for upgrade behavior from a healthy revision to a failing replacement revision, asserting correct rollout and readiness signals.
  * Updated the OLMv1 test payload ordering to run the new Boxcutter runtime progress-deadline scenarios earlier and adjust related scenario placement.

### DIFF
--- a/openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
+++ b/openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
@@ -1,0 +1,576 @@
+package specs
+
+import (
+	"archive/tar"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	buildv1 "github.com/openshift/api/build/v1"
+	"github.com/openshift/api/features"
+	imagev1 "github.com/openshift/api/image/v1"
+	"github.com/openshift/origin/test/extended/util/image"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	olmv1 "github.com/operator-framework/operator-controller/api/v1"
+
+	"github.com/openshift/operator-framework-operator-controller/openshift/tests-extension/pkg/env"
+	"github.com/openshift/operator-framework-operator-controller/openshift/tests-extension/pkg/helpers"
+	exutil "github.com/openshift/operator-framework-operator-controller/openshift/tests-extension/test/qe/util"
+)
+
+var _ = g.Describe("[sig-olmv1][Jira:OLM][OCPFeatureGate:NewOLMBoxCutterRuntime] clusterextension progress deadline", g.Label("NonHyperShiftHOST"), func() {
+	defer g.GinkgoRecover()
+
+	var oc = exutil.NewCLIWithoutNamespace("default")
+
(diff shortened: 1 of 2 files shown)


tmshort:
Needs a Jira ticket or `NO-ISSUE`

dtfranz:
/retitle OPRUN-4392,[OPRUN-4393](https://redhat.atlassian.net/browse/OPRUN-4393): Add OLMv1 progress deadline QE tests

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,576 @@
...
+	})
+
+	g.It("PolarionID:88331-[OTP]A ClusterExtension is created and a persistent error prevents it from being fully rolled out", func(ctx g.SpecContext) {
+		const caseID = "88331"
+		fixture := newRolloutFailureFixture(ctx, oc, caseID, []rolloutFailureBundle{
+			{Version: "1.0.2", ControllerImage: "wrong/image"},
+		})
+
+		g.By("creating a ClusterExtension with a 1-minute progress deadline for the failing bundle")
+		ce := fixture.newClusterExtension("test-ce-install-timeout-"+caseID, "1.0.2", "olm-sa", ptr.To(int32(1)))
```
dtfranz: How are you running the tests? I may be running this wrong - this step fails for me because the `progressDeadlineMinutes` must be `>= 10`. We solved that in the upstream tests by patching the CRD, but I don't see that happening in here.

dtfranz:
Don't think this will work but...

/test e2e-aws-ovn-techpreview-extended-f1

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: Following up - I opened a draft PR and tagged these tests as `ReleaseGate` so I could run them in the CI. They failed for the same reason they failed when I ran them myself - the test relies on the CRD already being modified to allow a value less than `10`. I also noticed that the equivalent tests are [skipped](https://github.com/openshift/operator-framework-operator-controller/blob/ecd140bbf6c4b244704e6c52a5bdd25007fb9030/openshift/Makefile#L39) in the upstream test run, likely because we are unable to make modifications to the CRD when tests run downstream.

I don't know if there's an alternative here apart from setting this to the actual minimum of `10` and doing the same test - but then we would need to introduce a 10 minute sleep to the test run adding a bunch of extra runtime.

bandrade: Addressed in e0817e50f. The test now uses progressDeadlineMinutes=10, matching the downstream CRD minimum, and OCP-88331 is marked [Slow] with a 15-minute condition timeout. The original-name metadata was preserved, and both make build-update and make verify-metadata pass.

bandrade:
Thanks for the context, Daniel. I reviewed #755 and agree that it makes sense to use it as the replacement, so I am closing this PR in favor of #755.

Please carry over the `[Slow]` classification for OCP-88331 and retain the ClusterExtension-level `ProgressDeadlineExceeded` assertion from the latest revision here. The current version of #755 validates that condition only on the ClusterObjectSet. I also noticed the remaining CodeRabbit template comment.

Once those are addressed, #755 looks good to me. Thanks for keeping me as a co-author.

PR #753, opened 2026-06-22
Do not merge: Ocp 88331 88332 automation manual check

dtfranz:
Exercising new tests manually since they won't run in the presubmits

## Summary by CodeRabbit

* **Tests**
  * Added an automated test suite validating OLMv1 ClusterExtension “progress deadline” behavior, including cases where bundle installation fails to meet the deadline and expected status conditions are surfaced.
  * Added upgrade scenario coverage that switches from healthy to failing bundles, confirming the new revision continues rolling out and reports probe failures and related status updates correctly.

PR #755, opened 2026-06-23
OPRUN-4392,OPRUN-4393: Add OLMv1 progress deadline QE tests + fixes

dtfranz:
Automate the ClusterExtension rollout failure coverage for OCP-88331 and OCP-88332 by building in-cluster bundle and catalog images for successful and failing bundle versions.

The new QE specs verify ProgressDeadlineExceeded on an initial failed rollout and ProbeFailure while upgrading to a bad revision under the BoxCutter runtime.

Supersedes #745 

Additional changes:
* Test respects the CRD minimum of 10 minutes for the timeout - upstream we would modify the CRD so we can do 1 minute, but we can't do that downstream.
* Fixed the `httpd` script

Test pass shown here: [PR](https://github.com/openshift/operator-framework-operator-controller/pull/753), [CI run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_operator-framework-operator-controller/753/pull-ci-openshift-operator-framework-operator-controller-main-e2e-aws-techpreview-olmv1-ext/2069332924829798400)

## Summary by CodeRabbit

* **Tests**
  * Added a gated end-to-end ClusterExtension spec validating `ProgressDeadlineMinutes`, including `ProgressDeadlineExceeded` and probe failure behavior, with coverage for persistent rollout issues across install and replace/upgrade scenarios.
  * Expanded the OLMv1 test payload with additional Boxcutter-runtime scenarios and related progress-deadline cases.
* **New Features**
  * Added a helper option to set `ProgressDeadlineMinutes` on ClusterExtension.
* **Chores**
  * Improved in-cluster bundle/image/catalog build and cleanup tooling with better defaults, templating, and exported resource helpers.
  * Updated embedded operator assets and CSV templating to align with configured bundle versions.


openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,576 @@
...
+}
+
+func replaceAll(input string, replacements map[string]string) string {
+	for old, newValue := range replacements {
+		input = strings.ReplaceAll(input, old, newValue)
+	}
+	return input
+}
+
+func expectClusterCatalogServing(ctx context.Context, name string) {
```
pedjak: **suggestion:** The file introduces `expectClusterCatalogServing`, `expectClusterExtensionCondition`, `expectClusterObjectSetCondition`, and `deleteObject` — all of which duplicate existing helpers:

- `expectClusterCatalogServing` → `helpers.ExpectCatalogToBeServing` (`pkg/helpers/catalogs.go:38`)
- `expectClusterExtensionCondition` → `olmv1util.ClusterExtensionDescription.WaitClusterExtensionCondition`
- `deleteObject` → cleanup functions returned by `helpers.CreateClusterCatalog`, `helpers.EnsureCleanupClusterExtension`, etc.

Reusing the existing helpers would cut ~34 lines and keep timeouts consistent across the test suite (these use hardcoded 3/5/12 min vs the established `DefaultTimeout`/`DefaultPolling` constants).

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,576 @@
...
+                configMap:
+                  name: test-script-{{ CASE_ID }}
+                  defaultMode: 0755
+              containers:
+              - name: manager
+                image: "{{ CONTROLLER_IMAGE }}"
+                command:
+                - /scripts/httpd.sh
+                ports:
+                - containerPort: 8080
```
pedjak: **suggestion:** `containerPort: 8080` but `httpd.sh` listens on 8081 (and all probes correctly target 8081). The field is informational-only so nothing breaks, but it's misleading if someone reads the CSV template later. Consider changing to 8081.

dtfranz: Fixed 👍

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: `deleteObject` is still used for a few things that don't appear to have relevant helpers - but I've otherwise cleaned things up by using some other available helpers for clusterextension and catalog.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+		exutil.SkipMicroshift(oc)
+		exutil.SkipNoOLMv1Core(oc)
+		helpers.RequireFeatureGateEnabled(features.FeatureGateNewOLMBoxCutterRuntime)
+		helpers.RequireImageRegistry(ctx)
+	})
+
+	g.AfterEach(func() {
+		if g.CurrentSpecReport().Failed() {
+			helpers.DescribeAllClusterCatalogs(context.Background())
+			helpers.DescribeAllClusterExtensions(context.Background(), "")
```
perdasilva: **Bug**: `helpers.DescribeAllClusterExtensions(context.Background(), "")` passes an empty namespace, which causes the function to return immediately (dump.go:88-90) — no ClusterExtension diagnostic output is ever printed on test failure.

All 12 other call sites in the codebase pass a real namespace. The fixture creates resources in a known namespace (e.g., `"ns-88331"`) but that value is not captured here.

Suggested fix: capture the fixture namespace and pass it:
```go
g.AfterEach(func() {
    if g.CurrentSpecReport().Failed() {
        helpers.DescribeAllClusterCatalogs(context.Background())
        helpers.DescribeAllClusterExtensions(context.Background(), fixture.Namespace)
    }
})
```
This requires restructuring so `fixture` is available in the AfterEach scope.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+
+func newRolloutFailureFixture(ctx g.SpecContext, oc *exutil.CLI, caseID string, bundles []rolloutFailureBundle) rolloutFailureFixture {
+	k8sClient := env.Get().K8sClient
+	namespace := "ns-" + caseID
+	catalogName := "test-catalog-" + caseID
+
+	g.By("creating a test namespace")
+	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
+	o.Expect(k8sClient.Create(ctx, ns)).To(o.Succeed(), "failed to create Namespace")
+	g.DeferCleanup(deleteObject, ns)
```
perdasilva: **Convention violation**: `DeferCleanup` is registered *after* `Create`, but the AGENTS.md convention says:
> Register defer cleanup BEFORE creating resources — Pattern: `defer resource.Delete(oc)` then `resource.Create(oc)` — Why: Ensures cleanup even if Create partially succeeds then fails

If `Create` succeeds server-side but returns a network error, `o.Expect` fails, `DeferCleanup` is never reached, and the Namespace leaks.

Same issue at lines 145, 173, 200 (RoleBinding, ImageStream, BuildConfig).

Suggested fix for each — register cleanup first:
```go
g.DeferCleanup(deleteObject, ns)
o.Expect(k8sClient.Create(ctx, ns)).To(o.Succeed(), "failed to create Namespace")
```

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+}
+
+func eventually(callback func(o.Gomega), timeout time.Duration) {
+	o.Eventually(callback).WithTimeout(timeout).WithPolling(helpers.DefaultPolling).Should(o.Succeed())
+}
+
+func deleteObject(obj client.Object) {
+	err := env.Get().K8sClient.Delete(context.Background(), obj)
+	if err != nil && !errors.IsNotFound(err) {
+		o.Expect(err).NotTo(o.HaveOccurred(), "failed to delete %s %s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName())
```
perdasilva: **Diagnostics issue**: `obj.GetObjectKind().GroupVersionKind().Kind` returns `""` for typed Go structs that haven't been through scheme conversion. The error message will read `"failed to delete  ns-88331"` (empty kind).

Also, calling `o.Expect(err).NotTo(...)` during cleanup can mask the original test failure. The dominant codebase pattern in `in_cluster_bundles.go` uses `_ = k8sClient.Delete(...)` to silently ignore cleanup errors.

Suggested fix — use `client.IgnoreNotFound` and drop the function:
```go
g.DeferCleanup(func() {
    _ = env.Get().K8sClient.Delete(context.Background(), obj)
})
```
Or if you want to keep the helper, use `client.IgnoreNotFound(err)` and log warnings instead of asserting.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+	g.DeferCleanup(cleanup)
+	helpers.ExpectCatalogToBeServing(ctx, catalogName)
+
+	return rolloutFailureFixture{
+		Namespace:   namespace,
+		CatalogName: catalogName,
+	}
+}
+
+func buildImage(ctx g.SpecContext, oc *exutil.CLI, namespace, name string, files map[string][]byte) {
```
perdasilva: **Reuse**: `buildImage`, `createBuildArchive`, namespace creation, and image-puller RoleBinding setup (lines 120-234) reimplement functionality already present in `pkg/helpers/in_cluster_bundles.go` (`createImageStream`, `createBuildConfig`, `startBuild`, `waitForBuildToFinish`, `createTempTarBall`, `createNamespace`, `createImagePullerRoleBinding`).

Notably, the existing `waitForBuildToFinish` dumps `oc get build -oyaml` and `oc logs build/... --tail=200` on failure (lines 324-331), which the new code does not — build failures here will produce opaque errors.

Consider exporting the existing helpers (or extracting a shared higher-level function) rather than maintaining a parallel implementation.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+
+		g.By("creating a ClusterExtension with a 10-minute progress deadline for the failing bundle")
+		ceName, cleanup := helpers.CreateClusterExtension("test", "1.0.2", fixture.Namespace, caseID,
+			helpers.WithCatalogNameSelector(fixture.CatalogName),
+			helpers.WithProgressDeadlineMinutes(10),
+		)
+		g.DeferCleanup(cleanup)
+
+		g.By("waiting for the first ClusterObjectSet to report ProgressDeadlineExceeded")
+		expectClusterObjectSetCondition(ctx, ceName+"-1", olmv1.TypeProgressing, metav1.ConditionFalse, olmv1.ReasonProgressDeadlineExceeded)
```
perdasilva: **Fragility**: ClusterObjectSet names are hardcoded as `ceName+"-1"` and `ceName+"-2"` (also at lines 98, 101, 104), coupling tightly to the controller's `fmt.Sprintf("%s-%d", ext.Name, revisionNumber)` naming convention.

The test already has `expectActiveRevisions` which reads names from `status.ActiveRevisions` — the same approach could discover COS names dynamically, making the tests resilient to future naming changes.

Suggested approach:
```go
revisions := getActiveRevisions(ctx, ceName)
o.Expect(revisions).To(o.HaveLen(1))
expectClusterObjectSetCondition(ctx, revisions[0], ...)
```

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+
+		actual := make([]string, 0, len(ce.Status.ActiveRevisions))
+		for _, revision := range ce.Status.ActiveRevisions {
+			actual = append(actual, revision.Name)
+		}
+		g.Expect(actual).To(o.ConsistOf(expected), "active revisions mismatch")
+	}, 3*time.Minute)
+}
+
+func eventually(callback func(o.Gomega), timeout time.Duration) {
```
perdasilva: **Missing context propagation**: The `eventually` wrapper does not call `.WithContext(ctx)`, so if the Ginkgo SpecContext deadline fires, the `Eventually` continues polling for the full timeout (up to 12 minutes) before failing.

All three callers receive a `ctx` parameter. Gomega supports `.WithContext(ctx)` for early cancellation.

Suggested fix:
```go
func eventually(ctx context.Context, callback func(o.Gomega), timeout time.Duration) {
    o.Eventually(callback).WithContext(ctx).WithTimeout(timeout).WithPolling(helpers.DefaultPolling).Should(o.Succeed())
}
```

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+		ceName, cleanup := helpers.CreateClusterExtension("test", "1.0.0", fixture.Namespace, caseID,
+			helpers.WithCatalogNameSelector(fixture.CatalogName),
+			func(ce *olmv1.ClusterExtension) {
+				ce.Spec.Source.Catalog.UpgradeConstraintPolicy = olmv1.UpgradeConstraintPolicySelfCertified
+			},
+		)
+		g.DeferCleanup(cleanup)
+
+		g.By("waiting for the initial installation to complete")
+		expectClusterExtensionCondition(ctx, ceName, olmv1.TypeInstalled, metav1.ConditionTrue, "", "")
```
perdasilva: nit: These two sequential `expectClusterExtensionCondition` calls re-fetch the same ClusterExtension object to check different conditions. A single `Eventually` block checking both `TypeInstalled=True` and `TypeProgressing=True/ReasonSucceeded` would halve API calls and avoid a 6-minute worst-case timeout (3+3). Same pattern at lines 101-104 with the COS (24-minute worst case: 12+12).

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+}
+
+func deleteObject(obj client.Object) {
+	err := env.Get().K8sClient.Delete(context.Background(), obj)
+	if err != nil && !errors.IsNotFound(err) {
+		o.Expect(err).NotTo(o.HaveOccurred(), "failed to delete %s %s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName())
+	}
+}
+
+const bundleDockerfile = `FROM scratch
```
perdasilva: nit: 130+ lines of inline YAML templates bypass the `go-bindata` pipeline (`pkg/bindata/`) used by other in-cluster-build tests. This means the YAML is never validated by `make bindata` / `make verify`, and can't be reused by future tests without copy-paste. Consider moving these to `pkg/bindata/qe/` for consistency.

perdasilva (commented):
Code review of the progress deadline QE tests. 9 findings below (ranked by severity).

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,508 @@
...
+			helpers.WithCatalogNameSelector(fixture.CatalogName),
+			helpers.WithProgressDeadlineMinutes(10),
+		)
+		g.DeferCleanup(cleanup)
+
+		g.By("waiting for the first ClusterObjectSet to report ProgressDeadlineExceeded")
+		expectClusterObjectSetCondition(ctx, ceName+"-1", olmv1.TypeProgressing, metav1.ConditionFalse, olmv1.ReasonProgressDeadlineExceeded)
+
+		g.By("verifying the ClusterExtension reports ProgressDeadlineExceeded")
+		expectClusterExtensionCondition(ctx, ceName, olmv1.TypeProgressing, metav1.ConditionFalse, olmv1.ReasonProgressDeadlineExceeded, "Revision has not rolled out for 10 minute(s).")
```
perdasilva: nit: The exact substring `"Revision has not rolled out for 10 minute(s)."` is tightly coupled to the controller's message formatting. If the wording changes even slightly (drops the period, changes pluralization), this test fails for a cosmetic reason. The `Reason` check (`ProgressDeadlineExceeded`) already validates the semantic behavior — consider dropping the message assertion or matching a looser substring like `"10 minute"`.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: Fixed in refactor.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: Fixed in refactor.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: `deleteObject` is gone - should no longer be an issue.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: Re-used as many helpers as I could - should be fixed.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: Fixed in refactor.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: Fixed in refactor.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: Fixed in refactor.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: All in-line bindata has been removed - fixed.

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
dtfranz: Fixed.

dtfranz:
@perdasilva @pedjak I believe I've addressed all your comments here - please take another look if you can!

openshift/tests-extension/test/qe/specs/olmv1_ce_progress_deadline.go
```diff
@@ -0,0 +1,249 @@
...
+	olmv1 "github.com/operator-framework/operator-controller/api/v1"
+
+	catalogbindata "github.com/openshift/operator-framework-operator-controller/openshift/tests-extension/pkg/bindata/catalog"
+	operatorbindata "github.com/openshift/operator-framework-operator-controller/openshift/tests-extension/pkg/bindata/operator"
+	"github.com/openshift/operator-framework-operator-controller/openshift/tests-extension/pkg/env"
+	"github.com/openshift/operator-framework-operator-controller/openshift/tests-extension/pkg/helpers"
+	exutil "github.com/openshift/operator-framework-operator-controller/openshift/tests-extension/test/qe/util"
+)
+
+var _ = g.Describe("[sig-olmv1][Jira:OLM][OCPFeatureGate:NewOLMBoxCutterRuntime] clusterextension progress deadline", g.Label("NonHyperShiftHOST"), func() {
```
perdasilva: The `[OCPFeatureGate:NewOLMBoxCutterRuntime]` tag should be in each `g.It` title, not here in `g.Describe`.

From the [QE README — Feature Gates section](https://github.com/openshift/operator-framework-operator-controller/blob/main/openshift/tests-extension/test/qe/README.md#optional-labels-in-migrationnew-test-cases-title):

> **Case 1: Test only runs when feature gate is enabled**
> - Add `[OCPFeatureGate:xxxx]` in **`g.It` title**

The suite-selection tooling filters on individual test names, so the tag needs to appear in each `g.It` title to take effect — placing it only in `g.Describe` means the gate filter is silently skipped for these two tests.

dtfranz:
Added `[Slow]` tag and `ProgressDeadlineExceeded` test for the `ClusterExtension` to 88331 per @bandrade 's comment [here](https://github.com/openshift/operator-framework-operator-controller/pull/745#issuecomment-4975985714).

grokspawn:
/lgtm
barring discovery during CI

dtfranz:
/retest
So far the techpreview run has failed for a different reason every single time, so I think in this case we're seeing some heavy instability.

tmshort:
/override ci/prow/openshift-e2e-aws-techpreview

> So far the techpreview run has failed for a different reason every single time, so I think in this case we're seeing some heavy instability.

Yes, I've been seeing it fail on other PRs in this repo too, with similar `[sig-node]` errors. I think we can override.

tmshort:
/label acknowledge-critical-fixes

Since this is just adding tests, I think we should allow it through.

tmshort:

Since this is just adding tests, I think we should allow it through, with correct label name

rashmigottipati (approved):
/lgtm

PR #765, opened 2026-07-16
OPRUN-4437: test: add allow-case for operator maxOCPVersion > cluster version

tmshort:
Add a second ReleaseGate-eligible OTE test verifying that an operator whose olm.maxOpenShiftVersion exceeds the current cluster version does not block cluster upgrade (InstalledOLMOperatorsUpgradeable stays True).

The existing test only covered the blocking path (maxOCPVersion == current version → False). This covers the complementary allow path (maxOCPVersion == next minor → True), directly exercising the normalization logic introduced for the 4.23/5.0 co-release boundary.

A nextMinorVersion() helper mirrors the 4.23→5.1 special case so the bundle annotation is always set to the correct next upgrade target.

## Summary by CodeRabbit

* **Tests**
  * Extended OLMv1 operator installation coverage to run against the next minor OpenShift version.
  * Added a scenario for the NewOLM feature gate to confirm operator installation does not block cluster upgrades when an operator’s maximum supported OpenShift version is higher than the current cluster version.
  * Continued to verify operators install successfully and remain upgradeable after installation.

### DIFF
--- a/openshift/tests-extension/.openshift-tests-extension/openshift_payload_olmv1.json
+++ b/openshift/tests-extension/.openshift-tests-extension/openshift_payload_olmv1.json
@@ -1498,6 +1498,16 @@
     "environmentSelector": {}
   },
   {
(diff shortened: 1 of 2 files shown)


tmshort:

This is adding a test-case.

PR #774, opened 2026-07-27
[release-4.21] OCPBUGS-99756: increase catalog HTTP client timeout from 10s to 5m

tmshort:
The 10-second timeout was too aggressive for large catalog responses. Increase it to 5 minutes to avoid timeouts when fetching catalogs.


tmshort:
This is the minimized version of #773. It excludes an HA fix that is likely not needed for release-4.21.

rashmigottipati (approved):
/lgtm
/approve 

/verified by @rashmigottipati

PR #775, opened 2026-07-27
OCPBUGS-89330: Fix cache path to avoid /var/cache/dnf conflict

ericahinkleRH:
## Problem

On OpenShift 4.19.32, operator-controller pods crash on startup with:
```
chmod /var/cache/dnf: operation not permitted
```

This happens because:
1. Default cache path is `/var/cache`
2. RHEL base image contains `/var/cache/dnf` owned by root
3. Pod runs as non-root (UID 1001)
4. `EnsureEmptyDirectory()` tries to chmod root-owned directories, which fails

## Solution

Change default cache path from `/var/cache` to `/var/cache/operator-controller` to avoid conflicts with system directories.

## Testing

- Build and deploy with new default
- Verify pod starts successfully
- Verify cache operations work correctly

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-89330

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default filesystem cache location for the operator-controller by changing the `--cache-path` default from `/var/cache` to `/var/cache/operator-controller`, keeping cache files in a dedicated directory for improved organization.

### DIFF
--- a/cmd/operator-controller/main.go
+++ b/cmd/operator-controller/main.go
@@ -181,7 +181,7 @@ func init() {
 	flags.BoolVar(&cfg.enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flags.StringVar(&cfg.cachePath, "cache-path", "/var/cache", "The local directory path used for filesystem based caching")
+	flags.StringVar(&cfg.cachePath, "cache-path", "/var/cache/operator-controller", "The local directory path used for filesystem based caching")
 	flags.StringVar(&cfg.systemNamespace, "system-namespace", "", "Configures the namespace that gets used to deploy system resources.")
 	flags.StringVar(&cfg.globalPullSecret, "global-pull-secret", "", "The <namespace>/<name> of the global pull secret that is going to be used to pull bundle images.")
 


rashmigottipati (commented):
Hi @ericahinkleRH, thank you for your contribution. 

I noticed the default `--cache-path` is defined upstream as `/var/cache`, and this change updates that default. Should this fix be made in the upstream operator-controller project first and then be pulled here through the sync? Or is this issue specific to the downstream OpenShift deployment? I'd like to understand whether there's a reason for carrying this as a downstream-only change.

ericahinkleRH:
Hi @rashmigottipati! This is OpenShift-specific because of the different base images:

Upstream uses `gcr.io/distroless/static:nonroot` which doesn't have `/var/cache/dnf`, while OpenShift uses `registry.ci.openshift.org/ocp/5.0:base-rhel9` which contains `/var/cache/dnf` owned by root.

When the pod runs as UID 1001, `EnsureEmptyDirectory()` tries to chmod `/var/cache/dnf` and fails with "operation not permitted" because non-root can't chmod root-owned directories.

Changing the default to `/var/cache/operator-controller` avoids this conflict with system directories in the RHEL base image while still working fine on distroless.

rashmigottipati (commented):
Thanks for the context @ericahinkleRH . That makes sense -the issue is tied to the OpenShift RHEL base image layout rather than the upstream default itself. Given that upstream's distroless image does not contain the conflicting `/var/cache/dnf` path, a downstream change here seems appropriate. lgtm from that perspective.

tmshort:
/hold
This can't be fixed here in main; as it's the upstream code.
If you want to fix it here in openshift, you can change the value of `-cache-path` in the manifest.

tmshort:
Given that we don't provide a mechanism in the upstream manifests to modify the arguments to the deployments; this should likely be done upstream.

ericahinkleRH:
Thanks for the explanation! You're absolutely right - I should have made this upstream first. 

I've created the upstream PR here: https://github.com/operator-framework/operator-controller/pull/2835

I included the OCPBUGS-89330 reference in the description so it will automatically attach to the downstreaming PR when it gets merged and pulled downstream.

Closing this PR in favor of the upstream fix. Once upstream merges and syncs down, the bug will be automatically attached to that PR.

Thanks for catching this!

PR #783, opened 2026-07-30
[release-4.22] OCPBUGS-100305: Fix cache path to avoid /var/cache/dnf conflict

ericahinkleRH:
Backport of #779 to release-4.22

**Jira**: https://redhat.atlassian.net/browse/OCPBUGS-100305

This fixes CrashLoopBackOff on startup due to `chmod /var/cache/dnf operation not permitted` by using dedicated cache directories instead of conflicting with system paths.


tmshort:
This can be verified via `/verified by CI` once the CI passes.

PR #794, opened 2026-08-31
OCPBUGS-104148: Replace golang.org/x/net with github.com/openshift-sustaining/net@v0.50.0-sec.3 to address CVE-2026-33814

sunku5494:


sunku5494:
/close
as this fix is already addressed in https://github.com/openshift/operator-framework-operator-controller/pull/801 for 4.21 branch

